PR #3, opened 2022-11-08
Answer:1 - Refactoring projection app from angular-challenges

matthieurom:
* Using content projection to display the correct image
* Adding event emitter so the add/remove store functions is not called in the UI component (`CardComponent`) anymore
* Adding `CityStore` and `CityComponent`
* Adding Item interface to get rid of `any[]` in `CardComponent`
* Remove `::ng-deep` by creating card.component.scss file


apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,18 @@
...
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px"
-  />
-
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
+  <ng-content></ng-content>
```
tomalaforge, line 4: nice, you can add a selector if you want to be more precise

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,18 @@
...
     <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname"
+      *ngFor="let item of items"
+      [name]="item.name"
       [id]="item.id"
-      [type]="type"
-    >
+      (_deleteItem)="_deleteItem.emit(item.id)">
     </app-list-item>
```
tomalaforge, line 11: What if I want to add an icon to my list-item? Or have a really custom item-list ?

Having only a name works for this very simple exemple, but we'll be very hard to highly customize in the future. 

You should look around ngTemplateOutlet

apps/projection/src/app/ui/card/card.component.scss
```diff
@@ -0,0 +1,11 @@
...
+  background-color: rgba(0, 250, 0, 0.1);
+}
+
+.bg-light-red {
+  background-color: rgba(250, 0, 0, 0.1);
+}
+
+.bg-light-blue {
+  background-color: rgba(0, 0, 250, 0.1);
+}
```
tomalaforge, line 11: that's a way of doing it. You can use css variable also. 
But for this example, you could only apply this class to the host, but applying class instead of an input.

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,18 @@
...
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Output() _addNewItem = new EventEmitter();
+  @Output() _deleteItem = new EventEmitter<number>();
```
tomalaforge, line 17: usually _ is reserved for private field (even if I personally don't like it), but I don't recommend using it for output.

apps/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,36 +1,49 @@
...
-    this.store.teachers$.subscribe((t) => (this.teachers = t));
+    this.store.teachers$
+      .pipe(
+        map((teachers) =>
+          teachers.map(
+            (t) => ({ name: `${t.lastname} ${t.firstname}`, id: t.id } as Item)
+          )
+        )
+      )
+      .subscribe((items) => (this.teacherItems = items));
```
tomalaforge, line 39: you should use the async pipe or at least you should unsuscribe from your observable somewhere. 

But using the async pipe let you simplify your code and setting your component to OnPush

tomalaforge:
Thanks for doing the first MR. Hope you will learn something.

apps/projection/src/app/ui/card/card.component.html
matthieurom: Selector added.

apps/projection/src/app/ui/card/card.component.html
matthieurom: I've added a ngTemplate inside `TeacherCardComponent`, `CityCardComponent`, and `StudentCardComponent` that will be used in `CardComponent` using content projection. Is this what you had in mind ?

apps/projection/src/app/ui/card/card.component.ts
matthieurom: Updated

apps/projection/src/app/component/teacher-card/teacher-card.component.ts
matthieurom: Yes you are right, I've added the modifications.

matthieurom:
> Thanks for doing the first MR. Hope you will learn something.

Thanks for this review, I've added the changes you suggested let me know what you think (particularly for the ngTemplateOutlet part). @tomalaforge

apps/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,52 @@
...
+    <ng-template #listItemTemplate let-item>{{ item?.name }} (C)</ng-template>
+  </app-card>`,
   standalone: true,
-  imports: [],
+  imports: [CardComponent, AsyncPipe],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CityCardComponent implements OnInit {
-  constructor() {}
+  cityItems$!: Observable<Item[]>;
```
tomalaforge, line 30: you can set your observable here instead of declaring it and setting it in ngOnInit. :)

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,25 @@
...
-  ) {}
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Input() listItemTemplate!: TemplateRef<any>;
```
tomalaforge, line 22: getting closer. 

You can user @ContentChild to get your template ref from your parent. 

and you should use listItemComponent in the parent : 
- you will not have an output going thought each component
- you can add ng-content inside this component and set whatever you want inside instead of concatening string inside name property.

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,21 @@
...
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
+  <ng-content select="[image]"></ng-content>
   <section>
     <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname"
+      *ngFor="let item of items"
+      [name]="item.name"
```
tomalaforge, line 8: not used anymore

apps/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,36 +1,57 @@
...
-      }
-    `,
-  ],
+    [listItemTemplate]="listItemTemplate"
+    (addNewItem)="addNewTeacher()"
+    (deleteItem)="deleteTeacher($event)">
+    <ng-container image>
+      <img src="assets/img/teacher.png" width="200px" />
+    </ng-container>
+    <ng-template #listItemTemplate let-item>{{ item?.name }} (T)</ng-template>
```
tomalaforge, line 27: nice 👍 
now you don't need to have a name property for each component, you can write whatever you want since it's projected.

And you should have you listItemcomponent here. everything will be simplier

tomalaforge:
getting really close. But you've got the point. Nice work

matthieurom:
> getting really close. But you've got the point. Nice work

Thanks for the reviews! It was fun!

apps/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,51 @@
...
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `<app-card
+    [items]="cityItems$ | async"
+    customClass="bg-light-blue"
+    (addNewItem)="addNewCity()">
+    <ng-container image>
+      <img src="assets/img/city.png" width="200px" />
+    </ng-container>
+    <ng-template #listItemRef let-item>
```
tomalaforge, line 22: this is it. Very nice. 
and just to be even more precise, You can call it let-city. 

And you can build your view from it and not in your component (which is not the place to do this)

<app-list-item>
  {{city.name}} {{city.country}}
</app-list-item>

and you can delete your map operaor l.34-37. 

cleaner 👍

PR #4, opened 2022-11-09
Answer:1 answer of projection challenge

tomalaforge:


AshishKumar409:
Hey! @tomalaforge , I don't think there is any use of card.component.html file as you are using template inside card.component.ts file. or could you please link the card.component.html file

tomalaforge:
Sorry, I didn't understand your question or remark. Can you be more specific?

AshishKumar409:
actually card.component.ts fils has its corresponding template file (card.component.html)  with html code, but still you used the template property inside the card.component.ts to write html.

tomalaforge:
Yes that's true. I updated my solution. I have forgotten to delete the file

PR #7, opened 2022-11-18
Answer:3 Ngrx selectors vs effect and reducers

tomalaforge:


acuevas91:
Is it the real solution for the challenge number 3 that refers to ngrx conceps? As I have seen so many changes referred to other newer challenges :)
Thanks in advance!

tomalaforge:
@acuevas91 thanks for pointing this out. I corrected the PR. Don't know what happened.

acuevas91:
> @acuevas91 thanks for pointing this out. I corrected the PR. Don't know what happened.

You're welcome! Thanks for your work"

PR #11, opened 2022-12-10
Answer:1 refactor content projection

arekgotfryd:


apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,22 @@
...
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px"
-  />
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
+  <ng-content select="[student]"></ng-content>
+  <ng-content select="[teacher]"></ng-content>
+  <ng-content select="[city]"></ng-content>
```
tomalaforge, line 6: the goal is to make a highly customizable component. That's the goal solution but why do you need to create one for each card ? 
If you need to add a new card, you will need to come here a add a new ng-content. Just create a global selector and this will perfect

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,22 @@
...
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item.firstname ? item.firstname : item.name"
       [id]="item.id"
-      [type]="type"
-    >
+      [type]="type">
     </app-list-item>
```
tomalaforge, line 14: having condition statement is not good for generic component. Everytime you will add a new card, you will have to check if the condition is still true or not. The goal is to create a very customizable list. 
You should go and learn about ngTemplateOutlet.

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -21,14 +26,17 @@ export class CardComponent {
...
   ) {}
 
   addNewItem() {
     if (this.type === CardType.TEACHER) {
       this.teacherStore.addOne(randTeacher());
     } else if (this.type === CardType.STUDENT) {
       this.studentStore.addOne(randStudent());
+    } else if (this.type === CardType.CITY) {
+      this.cityStore.addOne(randomCity());
     }
```
tomalaforge, line 40: This should not be here. Each store should stay in its specific component. 
use @Output() to emit data to your parent

tomalaforge:
Thanks for submitting your answer and trying the challenge. But the generic component is still to specific.

arekgotfryd:
Thanks for the review!
I've tried to incorporate all suggested changes. Could you please take a look?

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,36 @@
...
   selector: 'app-card',
   templateUrl: './card.component.html',
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
+  imports: [NgFor, ListItemComponent, CommonModule],
 })
 export class CardComponent {
-  @Input() list: any[] | null = null;
-  @Input() type!: CardType;
+  @Input() list!: Array<Teacher | City | Student>;
```
tomalaforge, line 21: You should not mention any specific data model. Use generic 

```typescript
export class CardComponent<T> {
@Input() list: T[]
}
```

typescript will infer the correct type depending on your list input

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,36 @@
...
   @Input() customClass = '';
+  @Input('optionTemplate')
+  optionTemplateRef!: TemplateRef<any>;
```
tomalaforge, line 24: yes that's one way to do it. 
You should add the read meta property to target the right element;
```typescript
@Input('optionTemplate', {read: TemplateRef})
  optionTemplateRef!: TemplateRef<any>;
```

Another way is by using contentChild. So, you will declare your template inside your parent selector and this will be easier to read, and to link that the template belong to that element. 

But more importanlty, don't use magic string like `optionTemplate` but use directive:
- you can add typing: [see this article](https://medium.com/@thomas.laforge/directive-type-checking-3d3c5f1137b3)
- no magic string

tomalaforge:
Nice. way better. 

Just added two additional comment to go from a good implementation to a great one.

arekgotfryd:
I added some suggested changes. Not sure if that is what you asked about but I gave it a try 😄

tomalaforge (approved):

PR #12, opened 2022-12-10
Answer:5

OmerGronich:
I tried adding tests with jest but failed in the configuration 🤔


tomalaforge:
Hello @OmerGronich, if you update the project and rebase into the newer branch, I added rxAngular, and now your test should work. 

I don't have lots of experience with rxAngular, so really happy to see your solution. Will try to check it and review the best I can.

apps/crud/src/app/app.component.ts
```diff
@@ -1,52 +1,115 @@
...
+        <button (click)="updatesSubject.next(todo)" [disabled]="todo.isLoading">
+          Update
+        </button>
+        <button
+          (click)="deleteSubject.next({todo, index})"
+          [disabled]="todo.isLoading">
+          X
+        </button>
+      </div>
+    </ng-template>
```
tomalaforge, line 40: Depending on what UX you want to offer, you can create a `todoItemComponent` to have a loader per item instead of a global one

PR #19, opened 2022-12-29
Answer:12. RxAngular solution

Karnaukhov-kh:


apps/scroll-cd/src/app/listen-outside-zone.util.ts
```diff
@@ -0,0 +1,16 @@
...
+): () => void {
+  getZoneUnPatchedApi(target, 'addEventListener').call(target, type, listener);
+  return () => {
+    getZoneUnPatchedApi(target, 'removeEventListener').call(
+      target,
+      type,
+      listener
+    );
+  };
+}
```
tomalaforge, line 16: question: why do you need to call getZoneUnPatchedApi twice ?

tomalaforge:
Thanks for submitting. 

Really neat solution. RxAngular is very powerful to do such thing.

apps/scroll-cd/src/app/listen-outside-zone.util.ts
Karnaukhov-kh: Thanks for accepting the solution :). `removeEventListener` is also patched by zone. So if you want to go all-in zone-less it's better to unpatch removal also.

Native 

remove-patched

Unpatched

remove-unpatched

tomalaforge: I'm so tired, I didn't see that this was two different listeners 😅

PR #21, opened 2022-12-29
Answer:8

tom-englert:

### DIFF
--- a/apps/pipe-easy/src/app/app.component.ts
+++ b/apps/pipe-easy/src/app/app.component.ts
@@ -1,21 +1,36 @@
 import { NgFor } from '@angular/common';
-import { Component } from '@angular/core';
+import {
+  ChangeDetectorRef,
+  Component,
+  Pipe,
+  PipeTransform,
+} from '@angular/core';
+
+@Pipe({
+  standalone: true,
+  pure: true,
+  name: 'heavyComputation',
+})
+export class HeavyComputationPipe implements PipeTransform {
+  transform(value: string, ...args: [index: number]) {
+    return `Piped@${new Date()}: ${value} - ${args[0]}`;
+  }
+}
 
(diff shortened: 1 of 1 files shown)


apps/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,36 @@
...
   `,
 })
 export class AppComponent {
   persons = ['toto', 'jack'];
 
-  heavyComputation(name: string, index: number) {
-    // very heavy computation
-    return `${name} - ${index}`;
+  constructor(cdr: ChangeDetectorRef) {
+    setInterval(() => cdr.markForCheck(), 1000);
```
tomalaforge, line 34: I think this is only for testing purposes ? 🧐

apps/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,36 @@
...
+  PipeTransform,
+} from '@angular/core';
+
+@Pipe({
+  standalone: true,
+  pure: true,
+  name: 'heavyComputation',
+})
+export class HeavyComputationPipe implements PipeTransform {
+  transform(value: string, ...args: [index: number]) {
```
tomalaforge, line 15: Why did you write `...args: [index: number]` this way ? 

and not only `index: number`

apps/pipe-easy/src/app/app.component.ts
tom-englert: Yes, just to see that the pipe output does not change

apps/pipe-easy/src/app/app.component.ts
tom-englert: You are right, that can be simplified.
I was already thinking about how to solve the typed challenge 😃

PR #22, opened 2022-12-29
Answer:9

tom-englert:


tomalaforge:
I corrected the readme file of this exercice because it was misleading. 

The goal is to create a highly reusable pipe, so you should not have your functions inside the pipe. Your functions should stay inside your component. The pipe will only wrap them so the result will be memoized.

tom-englert:
So like challenge #10, but without strong typing?

PR #24, opened 2022-12-29
Answer:12. fromEvent Solution

AlexScigalszky:

### DIFF
--- a/apps/scroll-cd/src/app/app.component.ts
+++ b/apps/scroll-cd/src/app/app.component.ts
@@ -1,6 +1,13 @@
-import { AsyncPipe, NgIf } from '@angular/common';
-import { Component, HostListener } from '@angular/core';
-import { BehaviorSubject } from 'rxjs';
+import { AsyncPipe, DOCUMENT, NgIf } from '@angular/common';
+import { Component, Inject } from '@angular/core';
+import {
+  distinctUntilChanged,
+  fromEvent,
+  map,
+  Subject,
+  takeUntil,
+  tap,
+} from 'rxjs';
 
 @Component({
   standalone: true,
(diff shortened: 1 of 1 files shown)


tomalaforge:
thanks for submitting your PR, but you changed the code but the CD is still running a lot. You should have only one CD when you want to display the button and one when you want to hide the button

AlexScigalszky:
Thanks for your review. I made some changes.
When I scroll down to bottom, just one change detection
image

Then If I scroll up to top, another charnge
image

My doubt here is when I click on the "to top" button
image
this call two change detection but I don't know how make it one.
NOTE: there three first are down, up and down again.

tomalaforge:
The solution is working fine now, great job

For your last screenshot: you cannot make it one without fine tuning zone js. 
- 1 cd for clicking on the button (zone js is listening to click event)
- 1 cd before you call window.scroll which is entering your function and you run `zone.run(() => display = false)`

AlexScigalszky:
Thanks you very much. It's nice learn something new every day

apps/scroll-cd/src/app/app.component.ts
```diff
@@ -31,16 +40,42 @@ import { BehaviorSubject } from 'rxjs';
...
+            () =>
+              window.pageYOffset ||
+              this.document.documentElement.pageYOffset ||
+              0
+          ),
+          map((pageYOffset: number) => pageYOffset > 50),
+          distinctUntilChanged(),
+          takeUntil(this.destroy$),
+          map((value: boolean) =>
+            this.zone.run(() => (this.displayButton = value))
```
LMFinney, line 70: The function in a `map` should be a pure function. This is using a map solely for a side effect. This behavior would belong in a `tap`, or even better the `subscribe` that you already have.

PR #25, opened 2022-12-30
Answer:8 

balaji-balamurugan:

### DIFF
--- a/apps/rxjs-pipe-bug/README.md
+++ b/apps/rxjs-pipe-bug/README.md
@@ -27,15 +27,15 @@ The QA team reports a **bug**. The UI shows **All [topic] have been deleted** al
 1. Fork the project
 2. clone it
 3. npm install
-4. **nx serve rxjs-pipe-bug**
+4. `npx nx serve rxjs-pipe-bug`
 5. _...work on it_
 6. Commit your work
 7. Submit a PR with a title beginning with **Answer:11** that I will review and other dev can review.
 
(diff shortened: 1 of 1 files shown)


apps/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,21 @@
...
+import { Component, Pipe, PipeTransform } from '@angular/core';
+@Pipe({
+  name: 'compute',
+  pure: true,
   standalone: true,
-  imports: [NgFor],
+})
+export class HeavyComputationPipe implements PipeTransform {
+  transform(value: { name: string; index: number }) {
+    return `${value?.name} - ${value?.index}`;
```
tomalaforge, line 10: never really though of passing my value this way, but this is working. Doing this way, if name of index change, both will be recompute. In our case this is what we want, so this is not an issue. But if you have static values you should consider doing like the following: 

the most common way is to have multiple property like 
```typescript
transform(name: string, index: number ) {
```

and in your html, you call your pipe like this:
```html
arg1 | mypipe : arg2
```

tomalaforge:
thanks for submitting your PR. nice work 👍

balaji-balamurugan:
thanks for marking it as the answer @tomalaforge

apps/pipe-easy/src/app/app.component.ts
balaji-balamurugan: yeah okay understood @tomalaforge

PR #26, opened 2022-12-30
Answer:1 Refactored things but found no solutions for better CSS (::part or CSS vars did not made it for me :))

liorchamla:


apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,42 @@
...
+        </app-list-item>
+      </section>
+
+      <button
+        class="border border-blue-500 bg-blue-300 p-2 rounded-sm"
+        (click)="add.emit()">
+        Add
+      </button>
+    </div>
+  `,
```
tomalaforge, line 32: This is working, but I think I should have ask for a more complex card to add. 

What if I want to add a text instead of a Title for my header ? I want to have a more highly customizable component. 
You should look at ng-content. 

Same for my list. What if I want to add a checkbox or another button ? In this case, I can only pass a name and id. 
You should look at ngTemplateOutlet. 

And having outputs going multiple components up, is not a very nice design, IMO.

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,42 @@
...
   standalone: true,
   imports: [NgIf, NgFor, ListItemComponent],
 })
 export class CardComponent {
-  @Input() list: any[] | null = null;
-  @Input() type!: CardType;
+  @Input() list: Item[] = [];
```
tomalaforge, line 37: a generic might be a better choice to be able to add any type of list

apps/projection/src/app/ui/card/card.component.ts
liorchamla: Ok so you want me to play with ng-content and ng-content-outlet .. this is maybe why the exercise is named Projection :D :D

tomalaforge:
For css variable

```typescript
@Component({
// ...
  styles: [
    `
      :host {
        --card-background: rgba(0, 250, 0, 0.1);
      }
    `,
  ],
// ....
})
export class StudentCardComponent {}

@Component({
// ...
  styles: [
    `
      :host {
        background-color: var(--card-background, blue);
      }
    `,
  ],
 // ...
})
export class CardComponent<T> {

```

apps/projection/src/app/ui/card/card.component.ts
tomalaforge: Exactly 🤣🤣🤣

PR #29, opened 2022-12-31
Answer:11

tom-englert:

### DIFF
--- a/apps/rxjs-pipe-bug/src/app/app.service.ts
+++ b/apps/rxjs-pipe-bug/src/app/app.service.ts
@@ -1,5 +1,5 @@
 import { inject, Injectable } from '@angular/core';
-import { merge, mergeMap, Observable, of, take } from 'rxjs';
+import { combineLatest, map, Observable, switchMap, take } from 'rxjs';
 import { LocalDBService, TopicType } from './localDB.service';
 
 @Injectable({ providedIn: 'root' })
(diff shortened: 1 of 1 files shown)


apps/rxjs-pipe-bug/src/app/app.service.ts
```diff
@@ -8,15 +8,13 @@ export class AppService {
...
       take(1),
-      mergeMap((topicToDelete) =>
-        topicToDelete.length > 0
-          ? topicToDelete
-              .map((t) => this.dbService.deleteOneTopic(t.id))
-              .reduce((acc, curr) => merge(acc, curr), of(true))
-          : of(true)
+      switchMap(topicsToDelete =>
+        forkJoin(topicsToDelete.map((t) => this.dbService.deleteOneTopic(t.id))).pipe(
+          map(results => !results.some(result => !result)))
```
tomalaforge, line 17: you can also use the `every` operator to avoid having a double negative condition

tomalaforge:
nice work

apps/rxjs-pipe-bug/src/app/app.service.ts
tom-englert: yes, meanwhile I found the docs that `every `on an empty array works as needed here.

apps/rxjs-pipe-bug/src/app/app.service.ts
```diff
@@ -8,15 +8,13 @@ export class AppService {
...
       take(1),
-      mergeMap((topicToDelete) =>
-        topicToDelete.length > 0
-          ? topicToDelete
-              .map((t) => this.dbService.deleteOneTopic(t.id))
-              .reduce((acc, curr) => merge(acc, curr), of(true))
-          : of(true)
+      switchMap(topicsToDelete =>
+        forkJoin(topicsToDelete.map((t) => this.dbService.deleteOneTopic(t.id))).pipe(
+          map(results => results.every(result => result)))
```
tomalaforge, line 17: 👍

PR #30, opened 2023-01-01
Answer:9

balaji-balamurugan:
solution for pipe intermediate challenges

created new pipe ( compute pipe ) and implemented it in app component template

breaking changes : none


apps/pipe-intermediate/src/app/app.component.ts
```diff
@@ -1,14 +1,41 @@
...
+  }
+
+  private isAllowed(age: number, isFirst: boolean) {
+    if (isFirst) {
+      return 'always allowed';
+    } else {
+      return age > 25 ? 'allowed' : 'declined';
+    }
+  }
+}
```
tomalaforge, line 29: I think my readme is still unclear. 

All functions need to stay inside the component. The pipe should only be a wrapper (a very generic pipe that you can use everywhere in any component without having to add a new function and condition to it)

balaji-balamurugan: I couldn't get what are you saying and yeah readme is always very clear, can you show me some examples?

tomalaforge: This exercice is to create a pipe to wrap a function to use the memoization property of a pure pipe. 

If you want to use this pipe on another component with different function, you will need to add those function inside the pipe and that's not what we want. 

You could only pass the callback of the function as params to your pipe

like 

```typescript
transform(value: any, func: Function) {
return func(value)
}
```

PR #32, opened 2023-01-02
Answer:10 Pipe-hard

balaji-balamurugan:
solution for pipe hard challenge

breaking change - none


apps/pipe-hard/src/app/app.component.ts
```diff
@@ -1,18 +1,42 @@
...
+  ): string {
+    if (funcName === 'isAllowed' && 'age' in value) {
+      return PersonUtils.isAllowed(value.age, value.isFirst, value.activityAge);
+    } else if ('name' in value) {
+      return PersonUtils.showName(value.name, value.index);
+    } else {
+      return '';
+    }
+  }
+}
```
tomalaforge, line 24: this pipe is not generic enough, if your helper file has 20 or more functions, this will become error-prone and less maintenable, because someone can forgot to add it here. 

You should have something similar to this 
```typescript
  transform(
    value: myType,
    functionName: myFuncType,
    ...args: myArgType
  ): MyReturnType{
    return PersonUtils[functionName](value, ...args);
  }
```
This way, you can add as many function inside your helper file without the need to change the pipe.

balaji-balamurugan: yeah, I got it, but I have doubts regarding the ...args parameters, 
It is fine when it has one to two args, if it has some five to six args, the template will look clumsy.
and then what is the value parameter doing here? I mean what role does the value parameter do?

tomalaforge: Yes I know that you are calling your pipe in a different way. You pass an object as first param and then the name of the function

for me I always pass the most important param as first parameters, so value is the first param, and ...args is if there is more. 

And you are right, your function should not have too many params, but you can have 1, 2 or 3. so ...args takes any number of parameters.

balaji-balamurugan: would you consider this as a correct answer? or should I need to refactor it?

tomalaforge: You should do what you want. Your answer is working fine but not very scalable and maintainable. It's up to you to improve it or leave it as it is. 😇

balaji-balamurugan: Pardon my ignorance, I couldn't see something that you are saying. can you please explain how it will not be scalable and maintainable?
where should I improve ?

PR #34, opened 2023-01-03
Answer: 2 - selectActivitiesWithTeachers

AlexScigalszky:


apps/ngrx-1/src/app/store/status/status.effects.ts
```diff
@@ -5,13 +5,14 @@ import { combineLatest, concatMap, map } from 'rxjs';
...
 import * as StatusActions from './status.actions';
+import * as AppActions from '../app.actions';
 import { Status } from './status.model';
 
 @Injectable()
 export class StatusEffects {
   loadStatuses$ = createEffect(() => {
     return this.actions$.pipe(
-      ofType(StatusActions.loadStatuses),
+      ofType(StatusActions.loadStatuses, AppActions.initStore),
```
tomalaforge, line 15: This effect should not exist. You are only transforming data already present inside the store => this should tell you that this should be inside a selector function. 

Effects are only great for doing asynchronous task, not transforming data. 

this is the primary goal of this challenge

apps/ngrx-1/src/app/store/activity/activity.selectors.ts
```diff
@@ -8,3 +10,17 @@ export const selectActivities = createSelector(
...
+  (activities, statusState) => {
+    return activities.map(
+      (activity) =>
+        ({
+          teachers: statusState.teachersMap.get(activity.type) ?? [],
+          ...activity,
+        } as ActivityWithTeachers)
+    );
+  }
+);
```
tomalaforge, line 26: nice, that's better.

PR #36, opened 2023-01-07
Answer:5 

balaji-balamurugan:
completed crud operation challenge using ngrx component store, added some styles to look good

breaking change: none


apps/crud/src/app/app.component.ts
```diff
@@ -1,52 +1,153 @@
...
+    todo: state.todo,
+    loading: true,
+    error: null,
+  }));
+
+  private handleError = this.updater((state, error: Error) => ({
+    todo: state.todo,
+    error,
+    loading: false,
+  }));
```
tomalaforge, line 110: you can use the spread operator to avoid writing: todo: state.todo

apps/crud/src/app/app.component.ts
```diff
@@ -1,52 +1,153 @@
...
+      tapResponse(
+        (todo) => {
+          this.setState({
+            todo,
+            loading: false,
+            error: null,
+          });
+        },
+        (error: Error) => {
+          this.handleError(error);
```
tomalaforge, line 94: warning: updateTodo is updating TodoItem inside this component but not inside the main todoList, your two items are not synchronised anymore. This can be error-prone.

tomalaforge:
great job, only one small mistake with the update function, otherwise everything is great.

apps/crud/src/app/app.component.ts
balaji-balamurugan: okay, forget that, thank you @tomalaforge

balaji-balamurugan:
Hi @tomalaforge I am waiting for your response, please take a look.

apps/crud/src/app/app.component.ts
```diff
@@ -1,52 +1,185 @@
...
-        }
+  @Input()
+  set todoItem(todo: Todo) {
+    this.patchState({
+      todo,
+      callState: 'LOADED',
+    });
+  }
+
+  readonly updateTodo$ = this.effect((todo: Observable<Todo>) => {
```
tomalaforge, line 94: just to be picky, this is a function, not an observable, so don't use the $ sign

tomalaforge:
This is better, the state is in sync. 

The two functions `updateTodo` and `deleteTodo` are a bit weird with no return result. This doesn't feel right. An output might be better to decoupled a bit everything. But it will get harder to do since you will need to set the callstate on each todo. See my [https://github.com/tomalaforge/angular-challenges/pull/15](answer with rx angular). Or you can look at [https://github.com/tomalaforge/angular-challenges/pull/27](answer with CS), but my two stores are strongly coupled. Don't know which one is better.

PR #37, opened 2023-01-07
Answer:13

balaji-balamurugan:
style challenge solution - removed all the inputs, added host context to style dynamically.

breaking changes: none


apps/styling/src/app/text.component.ts
```diff
@@ -1,16 +1,34 @@
...
+        color: var(--text-error);
+        font-size: 30px;
+      }
+
+      :host-context(.warning) {
+        color: var(--text-warning);
+        font-size: 25px;
+      }
+    `,
+  ],
```
tomalaforge, line 32: error and warning should be inside staticTextComponent. This component should be very generic.

.blue and .font-s-15 are way too specific. If you add new component, you will need to change this component again.

Your css variables are great but make use of them but styling your component from the outside.

Define all properties that can be customize with a default value (you can use a global file as your did, that's fine), but style from the outside.

balaji-balamurugan:
As you mentioned I change the style to staticTextComponent,
But I am not quite sure about styling of this text component, I don't know, I did correctly
If it is wrong, please guide me

balaji-balamurugan:
Yeah @tomalaforge Got it now, thanks for explaining ☺️

balaji-balamurugan:
Hi @tomalaforge, is that fine now?

tomalaforge:
Sorry, but that's not it. 😅
Maybe my exercice is too simple that you can trick it as you did. Because your solution is working. But if my `TextComponent` was a lot more complex, this won't work. 
The goal is to set Css var inside `TextComponent` on each property you want to customize from the outside. 
And on your main component, you can override these Css var with your custom value. 
Is it clearer ?

PR #39, opened 2023-01-17
Answer:1 Content projection

mirc1one:
Hello, I kinda did a full-on dynamic approach which in case you need to add more individual customization, it's gonna be costly 🥲


apps/projection/src/app/app.component.ts
```diff
@@ -1,18 +1,45 @@
...
+      type: CardType.STUDENT,
+      labelKey: 'firstname',
+      imageSrc: 'assets/img/student.webp',
+      bgColor: 'rgba(0, 250, 0, 0.1)',
+    },
+    {
+      type: CardType.CITY,
+      labelKey: 'name',
+    },
+  ];
```
tomalaforge, line 44: I'm not a big fan of this solution with static array of paramater. Not very easy to customize and all element must have the same shape. And you need to pass a type to know which store to hit.
Since Angular is a component based framework, it's always better to separate each element inside its own component. easier to customize at will. If you need to change one thing inside one element, you don't need to add if statement every where and you only need to impact one component.

tomalaforge:
Thanks for submitting your request 🙏

PR #47, opened 2023-02-16
Answer: 6

balaji-balamurugan:
updated hasRole directive and added new guard functions for canMatch guard and reRouting to access-denied page if no access

breaking change - none


tomalaforge:
Oh I'm so sorry. I haven't seen your PR. will check it tonigh or tomorrow

apps/permissions/src/app/has-role.directive.ts
```diff
@@ -1,68 +1,61 @@
...
+  }
 
-  ngOnDestroy(): void {
-    console.log('on destroy');
+  @Input()
+  set hasRoles(roles: Role[] | undefined) {
+    if (roles && roles?.length > 0) {
+      this.store.hasAnyRole(roles).subscribe((hasRole) => {
+        hasRole ? this.addTemplate() : this.addElseTemplate();
+      });
```
tomalaforge, line 30: missing the cleanup logic

apps/permissions/src/app/guards.ts
```diff
@@ -0,0 +1,27 @@
...
+      return isAdmin ? true : router.createUrlTree(['/access-denied']);
+    })
+  );
+};
+
+export const hasPermissions = (
+  roles: Role[],
+  userStore = inject(UserStore)
+) => {
+  const router = inject(Router);
```
tomalaforge, line 21: I think it's better to place every injectable service inside the parameter of your function (same as `userStore`) with a default value. 

- easier to test
- can pass services when used outside the injection context

apps/permissions/src/app/access-denied.component.ts
```diff
@@ -0,0 +1,20 @@
...
+import { RouterLink } from '@angular/router';
+
+@Component({
+  // eslint-disable-next-line @angular-eslint/component-selector
+  selector: 'access-denied',
+  standalone: true,
+  imports: [CommonModule, ButtonComponent, RouterLink],
+  template: `
+    <p class="text-red-900 text-7xl">Access Denied!</p>
+    <button app-button [routerLink]="['/']">Back to login</button>
```
tomalaforge, line 13: when passing a string. you don't need to bind your prop with [], you can simply write:
`routerLink='/'` 

But that's just a preference, I find it easier to read

PR #64, opened 2023-05-12
Answer:1

fcbento:

### DIFF
--- a/apps/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,38 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { CardComponent } from '../../ui/card/card.component';
+import { City } from '../../model/city.model';
+import { CardType } from '../../model/card.model';
+import { FakeHttpService } from '../../data-access/fake-http.service';
+import { CityStore } from '../../data-access/city.store';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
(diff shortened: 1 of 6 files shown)


apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -21,14 +26,17 @@ export class CardComponent {
 
   constructor(
     private teacherStore: TeacherStore,
-    private studentStore: StudentStore
+    private studentStore: StudentStore,
+    private cityStore: CityStore
```
tomalaforge, line 30: The goal of this exercice is to create a highly customizable component. No logic should be inside this component. 

you can look at ng-content and templateOutlet for projecting content

apps/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,38 @@
...
+
+  ngOnInit(): void {
+    this.fetchCities();
+  }
+
+  fetchCities(): void {
+    this.http.fetchCities$.subscribe((cities: City[]) =>
+      this.store.addAll(cities)
+    );
+    this.store.cities$.subscribe((cities: City[]) => (this.cities = cities));
```
tomalaforge, line 36: you should try to avoid to use the suscribe function. Try to always use asyncpipe or use a componentStore from ngrx or rxstore from rxangular. They will handle the subciption for you otherwise don't forget to unsuscribe.

And using subscribe will create imperative code which is less maintanable. Using asyncpipe or reactive pattern will create declarative code which is easier to understand and maintain since each variable are created at a unique place.

tomalaforge:
Thanks for submitting your work. sorry for the delay. I haven't seen your MR.

PR #71, opened 2023-06-29
Answer:14 rxjs race condition

AlDrag:
Observable is now subscribed to within the `TopicModalComponent` template instead of the AppComponent. This means the fetch only occurs on opening of the modal.

I decided to go with passing the observable through the `MAT_DIALOG_DATA`. An alternative would have been to just use the `TopicService` within the `TopicDialogComponent`. However that would involve a lot more code to allow it still be decoupled from the source. Although it's probably more flexible down the line.

### DIFF
--- a/apps/rxjs-race-condition/src/app/app.component.ts
+++ b/apps/rxjs-race-condition/src/app/app.component.ts
@@ -1,11 +1,6 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  inject,
-  OnInit,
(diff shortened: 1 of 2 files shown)


tomalaforge:
Thanks for your answer 👍

That's a way to do it. I've simplified the issue I met inside a client project because topics was used inside appcomponent. that's why it was done that way. Should have added the use of the topic variable inside AppComponent to make it a bit more trickier while passing the observable.

PR #73, opened 2023-06-29
Activity:2 ngrx-1

AlDrag:
_I'm pretty familiar with NgRx, I'm just not 100% sure my usages of actions are the best practice._

I removed the `Status` store in favour of deriving the state from selectors.

I removed `load` actions in favour of a single action in the app.component.ts indicating that it loaded. I took the `event` instead of `command` approach for naming my action. All effects then listen for this action to trigger their fetch requests and subsequent actions.

I created a `vm` (view model) selector in `app.selectors.ts` alongside its `selectTeachersByActivityMap` selector. Since these selectors were purely just for the `app.component.ts` view, I thought it made the most sense to have those selectors there. The `vm` selector is just responsible for producing a state that resembles exactly what the view is asking for.

As a minor adjustment, I also created a new `selectUserIsAdmin` selector to take advantage of memoization, although not needed in this example since the state doesn't change.


tomalaforge:
thanks for submitting this solution. no one has ever done it before since the challenge is a bit longer than others.

And that's a great solution. Everything should be going though selector to derive the state. 💪

Well done

apps/ngrx-1/src/app/app.selectors.ts
```diff
@@ -0,0 +1,34 @@
...
+          status.push(activity.teacher);
+        } else {
+          statusMap[activity.type] = [activity.teacher];
+        }
+        return statusMap;
+      }, {} as Record<ActivityType, Person[]>);
+    }
+    return {} as Record<ActivityType, Person[]>;
+  }
+);
```
tomalaforge, line 23: to avoir some nesting and make it more complicated to read, you should treat the case of notAdmin first
like 

```ts
if(!isAdmin) return blablabl

// case isAdmin 
```

try also to avoid 'as', it's better to type the return of the function

```ts
(isAdmin, actitivities):  Record<ActivityType, Person[]> => {
 return activities.reduce((statusMap:Record<ActivityType, Person[]>, activity) => {
 // do stuff 
}

}
```

apps/ngrx-1/src/app/app.component.ts
```diff
@@ -60,14 +50,11 @@ import { loadUsers } from './store/user/user.actions';
...
-    this.store.dispatch(loadUsers());
-    this.store.dispatch(loadStatuses());
+    this.store.dispatch(appLoaded());
   }
-
-  getAllTeachersForActivityType$ = (type: ActivityType) =>
-    this.store.select(selectAllTeachersByActivityType(type));
 }
+
+export const appLoaded = createAction('[AppComponent] Loaded');
```
tomalaforge, line 60: maybe in a different file might be better. 

if you are using Nx, in a seperate lib is better also because if the action is called at multiple place, you will get circular dependancies. (I should make a challenge about this as well, might be interesting)

apps/ngrx-1/src/app/app.selectors.ts
AlDrag: Nice idea with the early termination if statement.

I don't think it's possible to avoid the `as` type overrides? Because I need to type the `{}` initial value of the reducer. 

So I can simplify it to this:
```
(isAdmin, activities) => {
    if (!isAdmin) return {};

    return activities.reduce((statusMap, activity): Record<ActivityType, Person[]> => {
      const status = statusMap[activity.type];
      if (status) {
        status.push(activity.teacher);
      } else {
        statusMap[activity.type] = [activity.teacher];
      }
      return statusMap;
    }, {} as Record<ActivityType, Person[]>);
  }
```

Any ideas?

apps/ngrx-1/src/app/app.component.ts
AlDrag: What would you name the file in this case? This one is sort of a special case where it's about if the app loaded.

But what if you had an action that represents when a particular page loaded? e.g. `UsersComponent`. Should that action go in the `UsersComponent` like what I've done here? Or should it go into its own file? What should that file called?

apps/ngrx-1/src/app/app.selectors.ts
tomalaforge: maybe you are right. This is fine for me. 😊

apps/ngrx-1/src/app/app.component.ts
tomalaforge: I name it the same way as the component that dispatch it.
`app.actions.ts`

So if you create another appLoaded action inside another component. it will be inside `users.actions.ts`

What I like to do as well it, create a barrel object 

```ts
const appLoaded = createAction('[AppComponent] Loaded') // no export there

export const AppActions = { appLoaded} 
```

that way, it's very easy to see where the action is coming from when you use it: `store.dispatch(AppActions.appLoaded)`

It's a bit similar to use createActionGroup

AlDrag: Love it. Will make the change.

Would you do the same for selectors that are core to a component? e.g. like my `vm` selector approach. Maybe something like `app.selectors.ts`?

alan-bio:
@tomalaforge is the idea for this project to keep PRs indefinitely open?
(I don't mind either way, just want to make sure I'm not meant to close a PR or something).

AlDrag:
@tomalaforge is the idea for this project to keep PRs indefinitely open?
(I don't mind either way, just want to make sure I'm not meant to close a PR or something).

tomalaforge:
for now, I keep them open but if you want you can close it. It will still be visible when you want to see available solutions from a challenge readme file.

apps/ngrx-1/src/app/app.component.ts
AlDrag: @tomalaforge sorry to bring up this old thread.

I've been thinking about the separation of actions and reducers/effects in NgRx.

So I understand it's generally advised to have actions as `events` instead of `commands`. Such as `[AppComponent] loaded` instead of `[Users] Load Users`.

But going with the `commands` approach, that means your stores, such as the user store, needs to hold references to actions outside of it. Such as the user store listening to `[AppComponent] loaded`.

This goes against the old saying of Separation of Concerns etc (I think). But is this what NgRx generally recommends?

The only time I can see this being a problem is with feature modules that shouldn't be importing anything from core.

PR #77, opened 2023-07-06
Answer:1

navenduraisamy:

- added City card
- refactored student and teacher cards


apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -21,14 +26,17 @@ export class CardComponent {
```
tomalaforge, line 26: cartType should disappear. 

your component will get less scalable, maintainable everytime you add a new Type.

tomalaforge:
Thanks for trying my challenge. If you have any question, don't hesitate to ask them inside this thread, This can benefit other

PR #79, opened 2023-07-13
Answer: 30

simplifiedcourses:


apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -80,40 +91,88 @@ import { PhotoStore } from './photos.store';
...
+    // but it is to prove the concept of effects in the signalstate
+    effect(
+      () => {
+        if (this.select('search')) {
+          this.patch({ page: 1 });
+        }
+      },
+      { allowSignalWrites: true }
+    );
+  }
```
tomalaforge, line 161: I don't get what you are trying to prove ? 😅
But indeed a search function will be way easier to understand, and it will not need to write `allowSignalWrite` which indicate a bad pattern, IMO.

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -80,40 +91,88 @@ import { PhotoStore } from './photos.store';
...
+      this.selectMany(['search', 'page'])
+    ).pipe(
+      debounceTime(300),
+      tap(() => this.patch({ loading: true })),
+      switchMap(({ search, page }) =>
+        this.photoService
+          .searchPublicPhotos(search, page)
+          .pipe(finalize(() => this.patch({ loading: false })))
       )
     );
```
tomalaforge, line 137: error handling is missing

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -80,40 +91,88 @@ import { PhotoStore } from './photos.store';
...
-export default class PhotosComponent implements OnInit {
-  store = inject(PhotoStore);
-  readonly vm$ = this.store.vm$.pipe(
-    tap(({ search }) => {
-      if (!this.formInit) {
-        this.search.setValue(search);
-        this.formInit = true;
-      }
-    })
+export default class PhotosComponent extends SignalState<PhotoComponentState> {
```
tomalaforge, line 98: When do you decide you should create a "local" service instead of using inheritance ?

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -80,40 +91,88 @@ import { PhotoStore } from './photos.store';
...
-  search = new FormControl();
-
-  ngOnInit(): void {
-    this.store.search(
-      this.search.valueChanges.pipe(
-        skipWhile(() => !this.formInit),
-        debounceTime(300),
-        distinctUntilChanged()
+  constructor() {
+    super();
```
tomalaforge, line 117: you can get the localStorage here and initialize your state depending on the value. 

And effect is perfect for saving into the localstorage when search or page update, isn't it ?

apps/interop-rxjs-signal/src/app/list/photos.component.ts
simplifiedcourses: Agreed, It's needed here because it is an effect, but a function called in the template would be clearer and more declarative imo

apps/interop-rxjs-signal/src/app/list/photos.component.ts
simplifiedcourses: Good point, still need to do that

apps/interop-rxjs-signal/src/app/list/photos.component.ts
simplifiedcourses: In a component, I use inheritance for the state. If it would become so big that we need to split it up (composition) then I would extract the state as well. I just like to treat components as state machines

apps/interop-rxjs-signal/src/app/list/photos.component.ts
simplifiedcourses: That makes total sense

PR #81, opened 2023-07-21
Answer:31 migrate module to standalone

udha19:
Answer:31 migrate module to standalone


.vscode/settings.json
```diff
@@ -1,3 +1,23 @@
...
+    "statusBarItem.hoverBackground": "#89ac6b",
+    "statusBarItem.remoteBackground": "#a3be8c",
+    "statusBarItem.remoteForeground": "#15202b",
+    "titleBar.activeBackground": "#a3be8c",
+    "titleBar.activeForeground": "#15202b",
+    "titleBar.inactiveBackground": "#a3be8c99",
+    "titleBar.inactiveForeground": "#15202b99"
+  },
+  "peacock.color": "#a3be8c"
 }
```
tomalaforge, line 23: if you want to change vscode settings, do it on your own machine and not on the project settings, not to pollute your PR

tomalaforge:
Thanks for your PR. but your are missing a lot of files. You only migrate the AppModule, you are missing all RoutingModules. 

I let you finish

udha19:
thank you, sory for not read the statement and goal clearly, just too excited to start my new journey.

libs/module-to-standalone/shell/src/lib/main-shell.routes.ts
```diff
@@ -1,36 +1,45 @@
...
 
 export const appRoutes: Route[] = [
   { path: '', redirectTo: 'home', pathMatch: 'full' },
   {
     path: 'home',
-    loadChildren: () =>
+    loadComponent: () =>
       import('@angular-challenges/module-to-standalone/home').then(
-        (m) => m.ModuleToStandaloneHomeModule
+        (mod) => mod.HomeComponent
```
tomalaforge, line 11: just because I want to say something: `mod` is not the right word here. There is no more module. You should write `c` or `comp`. 

But otherwise that's perfect. Great job 💪

udha19: I see..
I will change it.

tomalaforge:
One last thing, if you want to use the `default` keyword to simplify your imports, you can, 

you can do:
`import { default } from "....."` and then you can simplify the import function.

PR #84, opened 2023-07-25
Answer:32 Switching from a function in a template to a variable to fix the navi…

Getsov:
…gation issue

### DIFF
--- a/apps/bug-cd/src/app/main-navigation.component.ts
+++ b/apps/bug-cd/src/app/main-navigation.component.ts
@@ -2,7 +2,7 @@ import { AsyncPipe, NgFor, NgIf } from '@angular/common';
 import { Component, Input, inject } from '@angular/core';
 import { RouterLink, RouterLinkActive } from '@angular/router';
 import { FakeServiceService } from './fake.service';
-import { tap } from 'rxjs';
+import { Observable, map, of } from 'rxjs';
 
 interface MenuItem {
   path: string;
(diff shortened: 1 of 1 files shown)


apps/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -43,20 +44,34 @@ export class NavigationComponent {
...
+  info$: any;
+  menus: MenuItem[] = [];
+  constructor() {
+    this.info$ = this.fakeBackend.getInfoFromBackend().pipe(
+      // Using tap operator to update menus via the function every time info$ changes
+      tap((info) => {
+        if (info !== null) {
+          this.menus = this.getMenu(info);
+        } else {
+          this.menus = this.getMenu('');
```
AlDrag, line 70: You should avoid relying on Angular's change detection and instead purely use an observable.

I think this wouldn't work if `changeDetectionStrategy` was set to `OnPush`. I could be wrong though 😅

apps/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -43,20 +44,34 @@ export class NavigationComponent {
...
+      <app-nav [menus]="menus" />
     </ng-template>
   `,
   host: {},
 })
 export class MainNavigationComponent {
   private fakeBackend = inject(FakeServiceService);
-
-  readonly info$ = this.fakeBackend.getInfoFromBackend();
+  info$: any;
```
AlDrag, line 61: Don't forget to use types 😃 

Should avoid using `any` in most cases if possible. Can use `unknown` if you don't actually know the type, but in this case, we do.

AlDrag (changes requested):
Just need to take advantage of RxJS more and it should be good!

tomalaforge:
Hi @Getsov, Thanks for submitting your first PR and answering first challenge. 

First that's great, you detected where the bug was coming from. Can you explain why we have an infinite loop here at render time?

And thanks @AlDrag for commenting the PR.

apps/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -34,7 +35,7 @@ interface MenuItem {
   },
 })
 export class NavigationComponent {
-  @Input() menus!: MenuItem[];
+  @Input() menus: MenuItem[] | undefined;
```
tomalaforge, line 38: We really want to have menu items here. 

We should write `@Input({required: true}) menu !: MenuItem[]`

But we can pass an empty array if no item in the menu

apps/bug-cd/src/app/main-navigation.component.ts
tomalaforge: as @AlDrag mentionned, always type your properties. avoid `any` or use it when you really have no other choice.

`unknown` is not a good type here as well. And `unknow` type are not easy to deal with.

apps/bug-cd/src/app/main-navigation.component.ts
tomalaforge: You coding imperativily, which means you properties are updated at multiple places inside your class. This way of coding feels more natural at first but very hard to debug and maintain.

prefer the declarative approach. meaning declare your properties at one place only. And using observable, those properties will react to any change inside your applications.

In practice:

Don't declare an empty menu, but transform your observable immediatly using the `map` operator to get a list of menuitems. 

Don't declare `info$: any` and assigning the prop later. Assign it immediatly

If you don't understand what I'm saying, you can contact me on linkedin to discuss those point together. I'll be happy to help you.

apps/bug-cd/src/app/main-navigation.component.ts
Getsov: You are right, needed for the type emitting, but figured out another approach

apps/bug-cd/src/app/main-navigation.component.ts
Getsov: Absolutely! Resolved, thanks!

apps/bug-cd/src/app/main-navigation.component.ts
Getsov: Now updated to declarative approach, thanks!

apps/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -40,25 +41,19 @@ export class NavigationComponent {
...
   host: {},
 })
 export class MainNavigationComponent {
   private fakeBackend = inject(FakeServiceService);
+  menus$!: Observable<MenuItem[]>;
+  constructor() {
+    this.menus$ = this.fakeBackend
+      .getInfoFromBackend()
+      .pipe(map((info) => this.getMenu(info || '')));
+  }
```
tomalaforge, line 54: This is better but why do you reassign your obs inside the constructor?

Assign it directly when you declare it.

Something you need to reassign it inside the ngoninit method because you need some inputs, but when you don't need any, no need to do it like it. 

it would be easier to read and you will not need to declare a type. Type will be infered by typescript directly

Getsov: Absolutely, forgot to move it when refactoring, now pushed

PR #85, opened 2023-07-26
Answer: 32 fix app not starting leveraging rxjs

Ostromecky:
## Explanation

My assumption: The `getMenu(...)` method initially has been called directly within template so it was being executed on every Change Detection Cycle. Additionally in the example we didn't leveraged `ChangeDetectionStrategy.OnPush` which means every DOM events, timers, promises, XHR, observables etc caused new Change Detection Cycle run. It leaded to much unnecessary method executions affecting app performance and finally to infinite loop. 

Changes affect behaviour as the component requests for menu info once on menu$ variable is being initialised and `menu` reference not changes on every cd cycle.

### DIFF
--- a/apps/bug-cd/src/app/main-navigation.component.ts
+++ b/apps/bug-cd/src/app/main-navigation.component.ts
@@ -1,6 +1,7 @@
 import { AsyncPipe, NgFor, NgIf } from '@angular/common';
 import { Component, Input, inject } from '@angular/core';
 import { RouterLink, RouterLinkActive } from '@angular/router';
+import { map } from 'rxjs';
 import { FakeServiceService } from './fake.service';
 
 interface MenuItem {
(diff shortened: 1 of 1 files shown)


tomalaforge:
Thanks for submitting your answer. And this is perfect 👍. How long did it take you to find the error?
And can you explain why this is happening?

tomalaforge:
Yes this is the global idea.

If we want to be more precise. RouterActiveLink execute `markForCheck` inside a new microtask which will create a new changedetection cycle and as you mentionned. the instance of the object returned by the function call inside the template is different, So Angular will update the view, which will delete and recreate two new rows inside the ngFor (using trackby can avoid the issue as well) and create a new instance of the RouterLinkActive and back to the beginning of the cycle.

Nice job 👍

Ostromecky:
Thanks for explanation :)

PR #86, opened 2023-07-26
Answer 1: projection

aramirezj:
Answer: 01

This is my take for the challenge 01, honestly I started coding only reading the readme and not the title xD, so my take its pretty different of the projection solution, but let me know what do you think please ^^, and very good repository of challenges!


tomalaforge:
Hi @aramirezj Thanks for submitting your solution

Very interesting take on this challenge. I'm not a big fan of callback functions. I find it harder to maintain and it is less scalable if you want to use the card component with a different use case. 
Moreover you still need to input a list with the same shape (name and id), How will you do if you want to add other properties or a button or anything else to your row? This is not very flexible

apps/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,47 @@
...
+  http: FakeHttpService = inject(FakeHttpService);
+  store: CityStore = inject(CityStore);
+  addRandom: () => City = randomCity;
+
+  ngOnInit(): void {
+    this.store.deleteOne;
+    this.http.fetchCities$.subscribe((c) => {
+      this.store.addAll(c);
+      this.store.citys$.subscribe((c) => (this.cities = c));
+    });
```
tomalaforge, line 45: nested subscribe should be avoided at all cost, 

prefer a declarative approach.

aramirezj:
> Hi @aramirezj Thanks for submitting your solution
> 
> Very interesting take on this challenge. I'm not a big fan of callback functions. I find it harder to maintain and it is less scalable if you want to use the card component with a different use case. Moreover you still need to input a list with the same shape (name and id), How will you do if you want to add other properties or a button or anything else to your row? This is not very flexible

Hi! Thank you! I agree, normally I only use callback function for very specific dummy components that want actually specific funcionality like create a custom button in a dummy component for anything. Normally for dummy component that are big I use a lot of Inputs and Output in order to pass all the information, for example a table component, all the model, columns, primary key, actions (in order to create the buttons).

Right now for the challenge wouldnt be very flexible you are right ^^ I think actually for this size of components the projection would be perfect, but do you agree for big components like the table I mentioned that could handle pagination, searching, ordering etc, would you go with an approach of Inputs outputs? Also in Angular 16 now with the transform has even potency, I would like to know your opinion.

And thank you for taking the time to answer!

tomalaforge:
When i say projection, I don't think of only using inputs ands output but also ng-content and NgTemplateOulter. The Dumb component will just handle the shape and some common operation like filtering/pagination/etc  and the smart component will handle the datas and how to display them using templateoutlet.

primeng/material or any component lib are using this approach. Otherwise their components will not be very flexibles.

I invite you to read my article on this subject to see what I'm trying to say to get a better idea. (you will find it on the readme file of the challenge)

PR #87, opened 2023-08-02
Answer:30 RxJS vs Signal Interop - Auto Signals

mfp22:
I implemented this RxJS/signal interop using a pattern I came up for this challenge: https://dev.to/mfp22/introducing-the-auto-signal-pattern-1a5h

### DIFF
--- a/apps/interop-rxjs-signal/src/app/inject-auto-signal.ts
+++ b/apps/interop-rxjs-signal/src/app/inject-auto-signal.ts
@@ -1,0 +1,12 @@
+import { DestroyRef, inject, ProviderToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+export function injectAutoSignal<
+  T,
+  Service extends { connection$: Observable<T> }
+>(token: ProviderToken<Service>) {
+  const service = inject(token);
+  const sub = service.connection$.subscribe();
+  inject(DestroyRef).onDestroy(() => sub.unsubscribe());
+  return service;
+}
--- a/apps/interop-rxjs-signal/src/app/list/photos.component.ts
+++ b/apps/interop-rxjs-signal/src/app/list/photos.component.ts
@@ -40,14 +40,14 @@ import { injectAutoSignal } from '../inject-auto-signal';
             [disabled]="store.page() === 1"
             [class.bg-gray-400]="store.page() === 1"
             class="text-xl border rounded-md p-3"
-            (click)="store.previousPage()">
+            (click)="store.previousPage$.next()">
             <
           </button>
           <button
             [disabled]="store.endOfPage()"
             [class.bg-gray-400]="store.endOfPage()"
             class="text-xl border rounded-md p-3"
-            (click)="store.nextPage()">
+            (click)="store.nextPage$.next()">
             >
           </button>
           Page :{{ store.page() }} / {{ store.pages() }}
--- a/apps/interop-rxjs-signal/src/app/list/photos.component.ts
+++ b/apps/interop-rxjs-signal/src/app/list/photos.component.ts
@@ -1,15 +1,13 @@
 import { NgFor, NgIf } from '@angular/common';
-import { Component, OnInit, inject } from '@angular/core';
-import { FormControl, ReactiveFormsModule } from '@angular/forms';
+import { Component } from '@angular/core';
+import { ReactiveFormsModule } from '@angular/forms';
 import { MatFormFieldModule } from '@angular/material/form-field';
 import { MatInputModule } from '@angular/material/input';
 import { MatProgressBarModule } from '@angular/material/progress-bar';
 import { RouterLinkWithHref } from '@angular/router';
-import { LetDirective } from '@ngrx/component';
-import { provideComponentStore } from '@ngrx/component-store';
-import { debounceTime, distinctUntilChanged, skipWhile, tap } from 'rxjs';
 import { Photo } from '../photo.model';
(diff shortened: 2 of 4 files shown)


apps/interop-rxjs-signal/src/app/list/photos.store.ts
```diff
@@ -30,106 +36,90 @@ const initialState: PhotoState = {
...
-      page: this.page$,
-      pages: this.pages$,
-      endOfPage: this.endOfPage$,
-      loading: this.loading$,
-      error: this.error$,
-    },
-    { debounce: true }
+  readonly search$ = this.searchControl.valueChanges.pipe(
+    startWith(this.searchControl.value),
+    filter((a) => !!a && a.length >= 3),
```
tomalaforge, line 68: you can write `filter((a): a is string => !!a && a.length >= 3)` to avoid having string | null

apps/interop-rxjs-signal/src/app/list/photos.store.ts
```diff
@@ -30,106 +36,90 @@ const initialState: PhotoState = {
...
-    } else {
-      const savedState = JSON.parse(savedJSONState);
-      this.setState({
-        ...initialState,
-        search: savedState.search,
-        page: savedState.page,
-      });
-    }
-  }
+  readonly searchResults$ = this.search$.pipe(
```
tomalaforge, line 73: you are not listening to page state change to fetch next page.

apps/interop-rxjs-signal/src/app/list/photos.store.ts
```diff
@@ -30,106 +36,90 @@ const initialState: PhotoState = {
...
+      const state = this.state();
+      if ('error' in result) {
+        return { ...state, error: JSON.stringify(result.error) };
+      } else {
+        return {
+          ...state,
+          photos: result.photos.photo,
+          pages: result.photos.pages,
+          error: '',
+        };
```
tomalaforge, line 105: Question: why don't you put this inside searchResults$ ?

apps/interop-rxjs-signal/src/app/list/photos.store.ts
```diff
@@ -30,106 +36,90 @@ const initialState: PhotoState = {
...
-  }
+  readonly searchResults$ = this.search$.pipe(
+    map((search) => ({ search: search || '', page: 1 })), // TypeScript thinks it's string | null
+    switchMap(({ search, page }) =>
+      this.photoService.searchPublicPhotos(search, page).pipe(
+        tap(() => {
+          localStorage.setItem(
+            PHOTO_STATE_KEY,
+            JSON.stringify({ search, page })
+          );
```
tomalaforge, line 81: I know you don't like effect, put localstorage is a good candidate for effect, no ? I would love your opinion on that ?

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -75,39 +72,17 @@ import { PhotoStore } from './photos.store';
...
-    this.store.search(
-      this.search.valueChanges.pipe(
-        skipWhile(() => !this.formInit),
-        debounceTime(300),
-        distinctUntilChanged()
-      )
-    );
-  }
+export default class PhotosComponent {
+  store = injectAutoSignal(PhotoStore);
```
tomalaforge, line 85: very nice, you are creating a viewmodel but without the ngIf 

And you do it this way to create a lazy subscription to signal ? Instead of using the store signal directly ? right ?

mfp22: Yes

apps/interop-rxjs-signal/src/app/list/photos.store.ts
mfp22: Doesn't effect just subscribe to the observable? This already was doing this with a tapResponse. Side-effects can happen inside taps. I just modified tapResponse, so I didn't really change this very much. I just moved it outside of effect and got something else to subscribe to it instead.

tomalaforge: I meant signal effect.

apps/interop-rxjs-signal/src/app/list/photos.store.ts
mfp22: Oh. Maybe I could have. I'll look at this more.

apps/interop-rxjs-signal/src/app/list/photos.store.ts
mfp22: I didn't think about that. I just modified the code from before. But maybe it wouldn't work there because at the end `injectAutoSignal` resets to initial state, and I wouldn't want an effect to override localStorage back to that.

apps/interop-rxjs-signal/src/app/inject-auto-signal.ts
```diff
@@ -0,0 +1,12 @@
+import { DestroyRef, inject, ProviderToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+export function injectAutoSignal<
+  T,
+  Service extends { connection$: Observable<T> }
+>(token: ProviderToken<Service>) {
+  const service = inject(token);
+  const sub = service.connection$.subscribe();
+  inject(DestroyRef).onDestroy(() => sub.unsubscribe());
```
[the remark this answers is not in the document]
mfp22: Thanks for this. I updated the article I wrote for this

PR #88, opened 2023-08-06
Answer:30 use @ngrx/signals for the PhotosStore impl

markostanimirovic:
`@ngrx/signals` playground: https://github.com/markostanimirovic/ngrx-signal-store-playground
RFC: https://github.com/ngrx/platform/discussions/3796


tomalaforge:
Thanks a lot for your work and your solution using ngrx signal store. 🙏 The solution is so nice. 

I will update the PR once the library has been released to better understand the PR. Because right now there is a lot of files. 😅

Thanks again

PR #95, opened 2023-08-29
Answer:30 solution using computedFrom

tomalaforge:


apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -32,88 +46,122 @@ import { PhotoStore } from './photos.store';
...
     return encodeURIComponent(JSON.stringify(photo));
   }
+
+  private initStore() {
+    const savedJSONState = localStorage.getItem(PHOTO_STATE_KEY);
+    if (savedJSONState === null) return;
+    const savedState = JSON.parse(savedJSONState);
+
+    this.searchForm.setValue(savedState.search);
+    this.search.set(savedState.search);
```
eneajaho, line 164:
```suggestion
```

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -32,88 +46,122 @@ import { PhotoStore } from './photos.store';
...
+    )
   );
 
-  private formInit = false;
-  search = new FormControl();
+  readonly photos = computed(() => this.photosAPI().photos.photo);
+  readonly pages = computed(() => this.photosAPI().photos.pages);
+
+  constructor() {
+    this.searchForm.valueChanges
```
eneajaho, line 134:
```suggestion
```

apps/interop-rxjs-signal/src/app/list/photos.component.ts
```diff
@@ -32,88 +46,122 @@ import { PhotoStore } from './photos.store';
...
+  searchForm = new FormControl();
+
+  readonly page = signal(1);
+  readonly search = signal('');
+  readonly loading = signal(false);
+  readonly error = signal('');
+  readonly endOfPage = computed(() => this.page() >= this.pages());
+
+  readonly photosAPI = computedFrom(
+    [this.search, this.page],
```
eneajaho, line 111:
```suggestion
    [this.searchForm.valueChanges, this.page],
```

eneajaho: Or what we can do is to extract the valueChanges to it's own variable with filtering, debouncing and other stuff. 

```ts
searchFormChanges$ =  this.searchForm.valueChanges.pipe(
  filter(([search]) => search.length >= 3),
  debounceTime(300),
  distinctUntilChanged(),
);
```

and use it as: 

```ts
readonly photosAPI = computedFrom(
  [this.searchFormChanges$, this.page],
  pipe(
    tap(() => this.loading.set(true)),
    switchMap(([search, page]) =>
      this.photoService.searchPublicPhotos(search, page).pipe(
        catchError((err: unknown) => {
          this.error.set(JSON.stringify(err));
          return EMPTY;
        }),
        finalize(() => this.loading.set(false))
      )
    ),
    startWith({ photos: { photo: [] as Photo[], pages: 0 } })
  )
);
```

This way we don't have to debounce the api call if the page changes, but we want it only when doing the search in the input.

tomalaforge: yes I thought of it during lunch break. I didn't like the subscribe and the distinctUntilChange with the custom function. 

And you are right about the debounceTime, it's not needed for page change.

Thanks for the review and nice work on the computedFrom. 👍

tomalaforge: I still need the search property as a signal because i listen to it inside my effect to save it inside the localStorage, but using `toSignal` is still nicer than using a subscribe function.

PR #96, opened 2023-09-01
Answer:3 Crud with component store/handelError/httpInterceptor

rd003:


apps/crud/src/app/todos/todo.store.ts
```diff
@@ -0,0 +1,117 @@
...
+  #deleteTodo = this.updater((state, id: number) => ({
+    ...state,
+    loading: false,
+    todos: state.todos.filter((t) => t.id !== id),
+  }));
+
+  // public methods
+  readonly loadTodos = this.effect<void>(
+    pipe(
+      tap((_) => this.#startLoading()),
```
tomalaforge, line 70: why do you use a underscore when there is no params? Just leave it empty 😉

apps/crud/src/app/todos/todo.store.ts
```diff
@@ -0,0 +1,117 @@
...
+  readonly vm$ = this.select(
+    {
+      todos: this.#todos$,
+      loading: this.#loading$,
+      error: this.#error$,
+    },
+    { debounce: true }
+  );
+
+  // private methods
```
tomalaforge, line 38: personal preference: I'm not a big fan of useless comment. 

using the # should be enough to know that the property is private.

tomalaforge:
Thanks for submitting your answer. very nice one. Great job 👍

rd003:
Thanks for reviewing. Yeah, underscore doesn't do anything, i put it out of
old habit.
I will definitely change it.

apps/crud/src/app/todos/todo.store.ts
rd003: It's kind of bad habit. I am aware that it _ doesn't add any value.
I usually type  _=> {...}
vs code convert it to (_)=> {....}

I will keep it in my mind.
Thanks for reviewing Thomas

PR #100, opened 2023-09-09
Answer:11 replace merge with combineLatest + map logic

georgegabor:
Answer:11


tomalaforge:
thanks for your solution. I haven't think of that solution, but this can be dangerous is obs doesn't complete on their own. 🙏

PR #101, opened 2023-09-09
Feature/answer 12 scroll cd

georgegabor:
Answer 12


tomalaforge:
Hi @georgegabor, thanks for answering my challenges. Can you separate both code. Inside this PR, you have some code related to another challenge. It would be nice to delete that code. Thanks 🙏

PR #102, opened 2023-09-18
Answer: 5 CRUD application

reed756:
Added update and delete buttons as well as NGrX component store and an item component. Also added tests to the todos service. Used spinners from Angular Material.

### DIFF
--- a/apps/crud/src/app/app.component.ts
+++ b/apps/crud/src/app/app.component.ts
@@ -1,6 +1,5 @@
 import { CommonModule } from '@angular/common';
 import { Component, OnInit } from '@angular/core';
-import { Todo } from './models/interface.todo';
 import { TodosService } from './services/todos.service';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { ComponentStore } from '@ngrx/component-store';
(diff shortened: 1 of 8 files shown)


apps/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,45 @@
...
-        }
-      )
-      .subscribe((todoUpdated: any) => {
-        this.todos[todoUpdated.id - 1] = todoUpdated;
-      });
+    this.todosService.initTodos().subscribe({
+      next: (todos: Todo[] | null) => {
+        this.todosStore.loadTodos(todos);
+      },
+    });
```
tomalaforge, line 43: Be careful with manual subscribe

Ngrx updater accept observable as input and they handle the subscription for you.

apps/crud/src/app/components/item.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  delete(todo: Todo) {
+    this.loadingText = 'Deleting...';
+    this.todosStore.startProcessing(todo.id);
+    this.todosService.deleteTodo(todo).subscribe(() => {
+      this.loadingText = '';
+      this.todosStore.stopProcessing();
+      this.todosStore.removeTodoFromList(todo.id);
+    });
+  }
+}
```
tomalaforge, line 57: all this code is imperative. Try to avoid this at best. This challenge has been created to practice declarative and reactive coding. 

You should always avoid setting properties inside a manual subscribe

Better way is to update your store and to listen to your local state and react to it. Not update the state and some properties inside your component. That's very dangerous because your state can be different than your local properties

apps/crud/src/app/services/todos.service.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    const req = httpMock.expectOne(
+      'https://jsonplaceholder.typicode.com/todos'
+    );
+    expect(req.request.method).toBe('GET');
+    req.flush(mockResponse);
+
+    service.todoList$.subscribe((response: Todo[]) => {
+      expect(response).toEqual(mockResponse);
+    });
+  });
```
tomalaforge, line 47: that's great to create test but be careful with testing observable. Try to always make your test fail before making it pass. 

I might be wrong but I think you can write anything inside your expectation (l.45) and your test will still pass. That's because your test ends before evaluating the assertion.

tomalaforge:
Thanks a lot for submitting your answer and trying this challenge. 

Try to do only declarative and reactive programming and try to delete all manual subscribe to practice. And I think you should rewrite your test. 

Otherwise nice job🙏

reed756:
Thanks for your review and also for creating the challenges! I'll have another go and get rid of the manual subscribes as well as redo the tests. :)

PR #103, opened 2023-09-22
Answer:28 Checkbox Testing

ti-he:

### DIFF
--- a/apps/testing-checkbox/src/app/app.component.spec.ts
+++ b/apps/testing-checkbox/src/app/app.component.spec.ts
@@ -1,10 +1,33 @@
 import { render } from '@testing-library/angular';
 import { AppComponent } from './app.component';
+import { screen } from '@testing-library/dom';
+import userEvent from '@testing-library/user-event';
+import { UserEvent } from '@testing-library/user-event/dist/types/setup/setup';
+import { ComponentFixture } from '@angular/core/testing';
 
 describe('AppComponent', () => {
+  let fixture: ComponentFixture<AppComponent>;
+  let user: UserEvent;
+
+  beforeEach(async () => {
+    fixture = (await render(AppComponent)).fixture;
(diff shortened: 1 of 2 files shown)


apps/testing-checkbox/src/app/app.component.spec.ts
```diff
@@ -1,10 +1,31 @@
...
+
+  beforeEach(async () => {
+    fixture = (await render(AppComponent)).fixture;
+    user = userEvent.setup();
+  });
+
   describe('When checking the checkbox', () => {
     it('Then button is enabled', async () => {
-      await render(AppComponent);
+      const checkbox = screen.getByTestId('checkbox') as HTMLInputElement;
```
tomalaforge, line 19: testId should be the last option to choose. Always prefer role or label or text before setting up a test-id

ti-he: Ah okay. I always thought testId's were commonly used. but I guess it makes sense to not clutter the html with them. Thanks for the quick review. 😄 

By the way, the [author solution](https://github.com/tomalaforge/angular-challenges/compare/main...solution-testing-checkbox) seems to be incorrect. The `user.click(checkbox)` has no `await` and is never executed properly. Yet the test passes, because it then checks if the button is disabled instead of enabled.

tomalaforge: testIf was a good practise when cypress came out. But testingLib changed the way we think. We should think more about a11y and getting our element with a11y parameters is always better than using a testId.

Thanks I"ll check my test 😅

tomalaforge: @ti-he Thanks for correcting the solution. I have no idea why I tested it the wrong way 😅

solution updated

PR #109, opened 2023-09-27
Answer: 34- Default onpush

jdegand:
Add an input component to reduce change detection calls.

### DIFF
--- a/apps/performance/default-onpush/src/app/person-list.component.ts
+++ b/apps/performance/default-onpush/src/app/person-list.component.ts
@@ -1,38 +1,17 @@
 import { Component, Input } from '@angular/core';
-
-import { CDFlashingDirective } from '@angular-challenges/shared/directives';
 import { CommonModule } from '@angular/common';
 import { FormsModule } from '@angular/forms';
 import { MatChipsModule } from '@angular/material/chips';
-import { MatFormFieldModule } from '@angular/material/form-field';
-import { MatInputModule } from '@angular/material/input';
 import { MatListModule } from '@angular/material/list';
 
 @Component({
   selector: 'app-person-list',
   standalone: true,
-  imports: [
-    CommonModule,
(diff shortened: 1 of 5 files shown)


apps/performance/default-onpush/src/app/person-list.component.ts
```diff
@@ -1,39 +1,18 @@
...
-  imports: [
-    CommonModule,
-    FormsModule,
-    MatListModule,
-    MatFormFieldModule,
-    MatInputModule,
-    MatChipsModule,
-    CDFlashingDirective,
-  ],
+  imports: [CommonModule, FormsModule, MatListModule, MatChipsModule],
```
tomalaforge, line 10: warning; you deleted the CDFlashingDirective. So you think the application is well optimized because CD is not flashing, but that's not the case.

apps/testing-router-outlet/src/app/app.component.spec.ts
```diff
@@ -1,31 +1,152 @@
...
+      await userEvent.click(button);
+
+      expect(
+        screen.queryByText(
+          'Borrowed Book: The Lord of the Rings by J.R.R. Tolkien'
+        )
+      ).toBeInTheDocument();
+      expect(
+        screen.queryByText('Borrowed Book: The Hobbit by J.R.R. Tolkien')
+      ).toBeInTheDocument();
```
tomalaforge, line 149: be careful, you need to create a new git branch and start from main branch again. Those commit should not be here.

tomalaforge:
Thanks for submitting your answer.

However your app is not optimized. When you type a new letter inside the input field, cd runs everywhere. Slitting your code inside a new component is a good start.

Can you create a new branch to separate your work from the different challenges? Thanks 🙏

If you have any questions, ask them here. I'll answer them

PR #110, opened 2023-09-27
Answer: 18 - testing nested

jdegand:
-  Cypress and Testing Library included.
-  Have [repo](https://github.com/jdegand/testing-nested) on my account with more info.

### DIFF
--- a/apps/testing-nested/src/app/child.component.cy.ts
+++ b/apps/testing-nested/src/app/child.component.cy.ts
@@ -4,15 +4,46 @@ describe('ChildComponent', () => {
   const setup = () => {
     cy.mount(ChildComponent);
   };
+
   describe('When typing nothing and clicking on Validate', () => {
     test('Then show "Title is required" error message and no http request has been sent', async () => {
       setup();
+
+      cy.window()
+        .its('console')
+        .then((console) => {
+          cy.spy(console, 'log').as('log');
+        });
+
+      cy.get(`[data-cy="input"]`).clear();
+
+      cy.get('button').click();
+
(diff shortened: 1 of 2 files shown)


apps/testing-nested/src/app/child.component.cy.ts
```diff
@@ -4,15 +4,46 @@ describe('ChildComponent', () => {
...
+
   describe('When typing nothing and clicking on Validate', () => {
     test('Then show "Title is required" error message and no http request has been sent', async () => {
       setup();
+
+      cy.window()
+        .its('console')
+        .then((console) => {
+          cy.spy(console, 'log').as('log');
+        });
```
tomalaforge, line 16: suggestion: Listening to console.log is not very clean. A better way is to mock your service and listen to the function call.

apps/testing-nested/src/app/child.component.spec.ts
```diff
@@ -1,16 +1,66 @@
...
+
+      const error = component.getByText('Title is required !!!');
+
+      expect(error).toBeInTheDocument();
+
+      const httpService = TestBed.inject<HttpService>(
+        HttpService
+      ) as Mock<HttpService>;
+
+      expect(httpService.sendTitle).not.toHaveBeenCalled(); //toHaveBeenCalledTimes(0)
```
tomalaforge, line 33: 👍 That's a better way to listen to a function call, why didn't you do that inside your cypress test?

tomalaforge:
Nice work. I have one suggestion but this is working.

PR #113, opened 2023-09-28
Answer:16 di

jdegand:
I was inspired by this [snippet](https://angularsnippets.dev/snippets/stateful-directive-with-output/) to create a directive with a store.  With this combination, you don't need the currency service or pipe.  

I also have an example of what changing the pipe (although not allowed) could look like in [my repo](https://github.com/jdegand/di).


tomalaforge:
Your solution is working, however the goal was to keep the pipe and the service and use the power of DI to get the right instance of the service inside the pipe.

apps/di/src/app/currency-code.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  @Input() set code(code: string) {
+    this.patchState({ code });
+  }
+
+  ngOnInit() {
+    this.code$.subscribe((symbol) => {
+      const ans = currency.find((c) => c.code === symbol);
+
+      this.el.nativeElement.insertAdjacentHTML('beforeend', `${ans?.symbol}`);
```
tomalaforge, line 40: If you are using a componentStore, you should leverage the method offered by the store. 
use effect in that case 👍
if you use a subscribe, don't forget to unsuscribe, otherwise you are creating a memory leak

tomalaforge, line 40: own opinion: I'm not a big fan of manipulating the DOM directly.

PR #117, opened 2023-09-28
Answer :8

BillyOkoth:
1. create a pipe to handle heavy computations on the template


apps/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,25 @@
...
     </div>
   `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class AppComponent {
   persons = ['toto', 'jack'];
 
   heavyComputation(name: string, index: number) {
     // very heavy computation
+    console.log('heavy computation');
```
tomalaforge, line 22: nitpick: console log need to be remove

PR #118, opened 2023-09-28
Answer:35

BillyOkoth:
1. creation of custom pipe to handle fibonacci computation


tomalaforge:
Thanks for your solution. Did you check the difference on the angular devtool ? I know that is pretty obvious but the goal of this serie is too learn how to detect bugs inside your own huge entreprise application. 

Nice work 👍

BillyOkoth:
Yes I looked into it it was awesome tool I had no knowledge about that thanks for the lesson from your challenges

PR #119, opened 2023-09-29
Answer:20 testing modal

jdegand:
I added data-cy attributes to the buttons because I had difficulty getting the click on the right part of the button. The buttons have multiple layers and targeting the inner layer prevents clicks from registering.

### DIFF
--- a/apps/testing-modal/src/app/app.component.cy.ts
+++ b/apps/testing-modal/src/app/app.component.cy.ts
@@ -32,9 +32,11 @@ describe(AppComponent.name, () => {
 
     cy.get('#mat-input-1').clear();
     cy.get('#mat-input-1').type('Bill');
-    cy.get('.mdc-button__label').click();
+    //cy.get('.mdc-button__label').click(); // this worked in this test
+    cy.contains('button', 'Confirm').click();
 
-    cy.get(`[data-cy="cancel"]`).click();
+    //cy.get(`[data-cy="cancel"]`).click();
+    cy.contains('button', 'Cancel').click();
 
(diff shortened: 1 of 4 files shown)


apps/testing-modal/src/app/app.component.cy.ts
```diff
@@ -1,19 +1,53 @@
...
   });
 
   test('confirm message is shown if you click "Confirm" in the confirmation modal after submitting a name', () => {
     setup();
+
+    cy.get('#mat-input-2').clear();
+    cy.get('#mat-input-2').type('Bill');
+    cy.get('.mdc-button__label').click();
+
+    cy.get(`[data-cy="confirmation"]`).click();
```
tomalaforge, line 49: `cy.contains('button', 'Confirm').click();` is working to get the button

apps/testing-modal/src/app/app.component.spec.ts
```diff
@@ -1,16 +1,77 @@
...
+
+    const input = screen.getByRole('textbox');
+
+    await userEvent.type(input, 'William');
+
+    await userEvent.click(button);
+
+    const cancelButton = document.querySelectorAll(
+      'button'
+    )[1] as HTMLButtonElement;
```
tomalaforge, line 43: this is dangerous, if you change position or add a new button, your test will fail which is a false position. (or you really want your test to fail because number of buttons and position matter)

Otherwise, you can do this: `    const confirmDialogButton = screen.getByRole('button', {
          name: /confirm/i,
        });`

PR #120, opened 2023-09-29
Answer:28 testing checkbox

jdegand:
Used fireEvent and userEvent.


tomalaforge:
that's a nice solution to display many ways of doing the same thing.

I just want to add that `userEvent` should be preferred over `fireevent` because it's closer to what a user would be able to do. 

For instance, if a button is not visible, fireevent will work but not userEvent because in real life, a user can not click on a invisible button.

PR #121, opened 2023-09-29
Answer:22 router input

jdegand:
I used transform to make the testId a number instead of a string.

I renamed permission to have a different name - you pass the name defined in the route to Input() and then you can change the variable to something else.


tomalaforge:
love the solution, shows all possibilities that the input API offer 👍

PR #122, opened 2023-09-29
Answer:19 testing input output

jdegand:
I tried to use componentProperties for the output test but I don't think it is possible.


apps/testing-input-output/src/app/counter.component.spec.ts
```diff
@@ -1,25 +1,105 @@
...
-        await render(CounterComponent);
+        await render(CounterComponent, {
+          componentProperties: { initialValue: 10 },
+        });
+
+        const counter = screen.getByTestId('counter');
+
+        const button = screen.getAllByRole('button')[0]; //increment is first button
+
+        // need to look up -> any way to do multiple clicks at once?
```
tomalaforge, line 27: don't think so. The test reproduce real life event and user can only click once 😇

PR #123, opened 2023-09-29
Answer:1 projection challenge

pdarko99:
…ut modifications

### DIFF
--- a/apps/projection/src/app/ui/card/card.component.html
+++ b/apps/projection/src/app/ui/card/card.component.html
@@ -17,7 +17,7 @@
     <ng-container *ngFor="let item of list">
       <ng-template
         [ngTemplateOutlet]="cardList"
-        [ngTemplateOutletContext]="{ item }">
+        [ngTemplateOutletContext]="{ $implicit: item }">
       </ng-template>
     </ng-container>
   </section>
--- a/apps/projection/src/app/ui/card/card.component.html
+++ b/apps/projection/src/app/ui/card/card.component.html
@@ -1,29 +1,30 @@
-<div class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3" [class]="customClass">
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [ngClass]="{
+    'bg-light-red': isTeacher,
+    'bg-light-green': isStudent
+  }">
   <img
     *ngIf="type === CardType.TEACHER"
     src="assets/img/teacher.png"
-    width="200px"
-  />
+    width="200px" />
   <img
     *ngIf="type === CardType.STUDENT"
     src="assets/img/student.webp"
-    width="200px"
-  />
+    width="200px" />
 
   <section>
-    <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname"
-      [id]="item.id"
-      [type]="type"
-    >
-    </app-list-item>
+    <ng-container *ngFor="let item of list">
+      <ng-template
+        [ngTemplateOutlet]="cardList"
+        [ngTemplateOutletContext]="{ item }">
+      </ng-template>
+    </ng-container>
   </section>
 
   <button
     class="border border-blue-500 bg-blue-300 p-2 rounded-sm"
-    (click)="addNewItem()"
-  >
+    (click)="addNewItem()">
     Add
   </button>
 </div>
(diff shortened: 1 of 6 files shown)


apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,30 @@
...
-      *ngFor="let item of list"
-      [name]="item.firstname"
-      [id]="item.id"
-      [type]="type"
-    >
-    </app-list-item>
+    <ng-container *ngFor="let item of list">
+      <ng-template
+        [ngTemplateOutlet]="cardList"
+        [ngTemplateOutletContext]="{ item }">
```
tomalaforge, line 20: you can use the $implicit property for the first value.

This way you will only use `let-xxx` inside your template instead of `let-xxx=item`

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,21 +1,46 @@
...
+      }
+    `,
+  ],
 })
 export class CardComponent {
+  @Input() isTeacher = false;
+  @Input() isStudent = false;
+
+  @ContentChild(CardListDirective, { read: TemplateRef })
+  cardList!: TemplateRef<any>;
```
tomalaforge, line 40: you can type the Template instead of using any.

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,21 +1,46 @@
...
+  TemplateRef,
+} from '@angular/core';
 import { randStudent, randTeacher } from '../../data-access/fake-http.service';
 import { StudentStore } from '../../data-access/student.store';
 import { TeacherStore } from '../../data-access/teacher.store';
 import { CardType } from '../../model/card.model';
-import { ListItemComponent } from '../list-item/list-item.component';
+
+@Directive({ selector: 'ng-template[card-list]', standalone: true })
+export class CardListDirective {}
```
tomalaforge, line 16: very nice 👍 using a directive instead of a magic string is awesome 💪

apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,34 @@
...
 })
 export class ListItemComponent {
   @Input() id!: number;
   @Input() name!: string;
   @Input() type!: CardType;
+  @Output() deleteEvent = new EventEmitter<number>();
 
   constructor(
     private teacherStore: TeacherStore,
     private studentStore: StudentStore
```
tomalaforge, line 28: you can delete those

apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,34 @@
...
 @Component({
   selector: 'app-list-item',
   template: `
+    <!-- <ng-template #appListItemTemplate let-item="item"> -->
     <div class="border border-grey-300 py-1 px-2 flex justify-between">
       {{ name }}
       <button (click)="delete(id)">
         <img class="h-5" src="assets/svg/trash.svg" />
       </button>
     </div>
```
tomalaforge, line 15: ng-content will create more flexibility to your component

tomalaforge:
very nice work, I added some non blocking comment.

apps/projection/src/app/ui/list-item/list-item.component.ts
pdarko99: hello i dont get it, if i delete it how am i going to delete ?.. That's what am using to delete an item

apps/projection/src/app/ui/list-item/list-item.component.ts
pdarko99: yeah seen :) thanks 👍

apps/projection/src/app/ui/card/card.component.ts
pdarko99: please can you elaborate more on this for me?

tomalaforge: You can add an interface instead of any to tell what context the user need to set

apps/projection/src/app/ui/list-item/list-item.component.ts
tomalaforge: for me, list item is just a container with a delete button, the item doesn't need to have any information. The parent can hold the information. So the parent will listen to the delete event and give the correct information at that time (in the parent).

pdarko99: Made changes .. thanks very much!

tomalaforge: that's way better 👍

pdarko99: thank you 👍

PR #125, opened 2023-09-29
Answer:1 projection challenge

alannelucq:


apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,17 @@
...
+import { Component, EventEmitter, Output } from '@angular/core';
 
 @Component({
   selector: 'app-list-item',
   template: `
     <div class="border border-grey-300 py-1 px-2 flex justify-between">
-      {{ name }}
-      <button (click)="delete(id)">
+      <ng-content />
+      <button (click)="deleted.emit()">
```
tomalaforge, line 8: awesome 👍

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,42 @@
...
+  `,
+  standalone: true,
+  host: {
+    class: 'border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3',
+  },
+  imports: [NgIf, NgFor, NgTemplateOutlet],
+})
+export class CardComponent<T> {
+  @Input({ required: true }) list: T[] = [];
+  @ContentChild('row', { read: TemplateRef }) row: TemplateRef<any>;
```
tomalaforge, line 40: Minor: to avoid magic string, you can use a directive instead of `row`

Minor: You can also type your template, instead of using any.

tomalaforge:
Great work 💪

apps/projection/src/app/ui/card/card.component.ts
alannelucq: It's fixed, thanks for your feedback 😁

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,51 @@
...
+@Directive({
+  selector: 'ng-template [card-list-item]',
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
 })
-export class CardComponent {
-  @Input() list: any[] | null = null;
-  @Input() type!: CardType;
-  @Input() customClass = '';
+export class CardListItemDirective {}
```
tomalaforge, line 16: 👍

tomalaforge:
That's perfect 🚀

PR #127, opened 2023-09-29
Answer: 34: Default vs OnPush

tomer953:
Ok, my first attempt, and I chose to start with this easy one, just to test myself.
I love your hard work. Big shout-out!

b.t.w The CDFlashingDirective is very cool directive, I guess we can embrace this idea for creating a testing\debugging tool even for real projects (and maybe add this to [ngxtension)](https://ngxtension.netlify.app/)


tomalaforge:
Nice idea for putting the CDFlashingDirective inside ngxtension. I'll do it 🙏

PR #128, opened 2023-09-29
Answer:35 Memoization

tomer953:
Oh, pipes. One day we'll miss them.
_Ahem, computed(), ahem._

### DIFF
--- a/apps/performance/memoized/src/app/fibonacci.pipe.ts
+++ b/apps/performance/memoized/src/app/fibonacci.pipe.ts
@@ -1,0 +1,30 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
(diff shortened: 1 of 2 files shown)


apps/performance/memoized/src/app/fibonacci.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+    c = 0;
+
+  for (let i = 3; i <= num; i++) {
+    c = a + b;
+    a = b;
+    b = c;
+  }
+
+  return c;
+};
```
tomalaforge, line 30: 😅 thanks

tomalaforge:
Yes we'll miss them, but that time are not arrived yet. Before all entreprise projects move to v16 at least to enjoy signals, it is not tomorrow

PR #129, opened 2023-09-29
Answer:1 projection

marryday:


apps/projection/src/app/directive/card-row/card-row.directive.ts
```diff
@@ -0,0 +1,19 @@
...
+})
+export class CardRowDirective<T extends object> {
+  @Input({ required: true }) appCardRow!: T[];
+
+  static ngTemplateContextGuard<TContext extends object>(
+    directive: CardRowDirective<TContext>,
+    context: unknown
+  ): context is Row<TContext> {
+    return true;
+  }
```
tomalaforge, line 18: 😍

apps/projection/src/app/service/destroy.service.ts
```diff
@@ -0,0 +1,14 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { ReplaySubject } from 'rxjs';
+
+@Injectable()
+export class DestroyService extends ReplaySubject<void> implements OnDestroy {
```
tomalaforge, line 5: question: why do you use a replaySubject and not a subject ?

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,24 @@
...
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+export class CardComponent<T> {
+  @Input({ required: true }) list: T[] = [];
+  @ContentChild(CardRowDirective, { read: TemplateRef }) row:
+    | TemplateRef<any>
+    | undefined;
```
tomalaforge, line 22: Minor: you could type your template instead of using `any`.

apps/projection/src/app/component/student-card/student-card.component.ts
```diff
@@ -1,36 +1,63 @@
...
 
-  constructor(private http: FakeHttpService, private store: StudentStore) {}
+  students = toSignal(this.store.students$, {
+    initialValue: [] as Student[],
+  });
 
   ngOnInit(): void {
-    this.http.fetchStudents$.subscribe((s) => this.store.addAll(s));
+    this.http.fetchStudents$
+      .pipe(takeUntil(this.destroy$))
```
tomalaforge, line 52: instead of using your own implentation of the DestroyService, angular now provide `takeUntilDestroyed()`

You can pass an injector into it, if run outside of the injector context like in your case.

tomalaforge:
By far, the most complete solution. You even rewite it into signal. Great great work. 💪⭐️

apps/projection/src/app/component/student-card/student-card.component.ts
marryday: True, I forgot about that

apps/projection/src/app/service/destroy.service.ts
marryday: I used to use just a regular subject, but later I've found taiga-ui's [implementation](https://github.com/taiga-family/taiga-ui/blob/main/projects/cdk/services/destroy.service.ts
) where they use ReplaySubject and left a comment 

```
 * Why we use `ReplaySubject` instead of `Subject`?
 * Well, we’ll use ReplaySubject to emit the last message in case
 * the subscription is ended after the component is destroyed.
```
But tbh I still don't get why memory leak should happen in that case.

apps/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,59 @@
...
+      <app-list-item (delete)="onDeleteItem(city.id)">
+        <div>{{ city.country }}</div>
+        <div>{{ city.name }}</div>
+      </app-list-item>
+    </ng-template>
+  </app-card>`,
   standalone: true,
-  imports: [],
+  providers: [DestroyService],
+  imports: [
```
Wanddoze, line 28: The imports is Duplicated!

PR #131, opened 2023-09-30
Angular: 8 Pure Pipe

DeveshChau:


tomalaforge:
Can you provide two separate PR for both challenges, thanks ?

PR #133, opened 2023-09-30
Answer:15 Function Overload

tomer953:
I don't really like this: `fuelOrCapacity?: Fuel | number,`  approach
but it solved the challange I guess


tomalaforge:
I don't like it either. But it was only for the purpose of the challenge. At that time, I didn't find a better idea 😅

PR #140, opened 2023-09-30
Answer:5

HosseinSalmanian:
Crud app refactored for pulling local state out of component using component store. Some other features added such as a global loading indicator for fetching todos list, local loading indicator for each todo item and an error message for each crud action which is performed on a todo item. some test also are wrote for todo store service.

### DIFF
--- a/apps/crud/src/app/services/todo.service.ts
+++ b/apps/crud/src/app/services/todo.service.ts
@@ -1,0 +1,170 @@
+import { Injectable } from '@angular/core';
+import { ComponentStore } from '@ngrx/component-store';
+import { TodoItem } from '../models/todo.model';
+import { TodoApiService } from './todo-api.service';
+import {
+  catchError,
+  concatMap,
+  finalize,
+  map,
+  switchMap,
+  tap,
+} from 'rxjs/operators';
+import { Observable, of } from 'rxjs';
+
+interface TodoStoreState {
+  todos: TodoItem[];
+  isLoading: boolean;
+}
+
+const initialState: TodoStoreState = {
+  todos: [],
+  isLoading: false,
+};
+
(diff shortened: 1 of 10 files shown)


apps/crud/src/app/services/todo.service.ts
```diff
@@ -0,0 +1,170 @@
...
+  public viewModel$ = this.select(
+    this.$todos,
+    this.$isLoading,
+    (todos, isLoading) => {
+      return {
+        todos,
+        isLoading,
+      };
+    }
+  );
```
tomalaforge, line 39: there is a new API where you can directly write : 
```ts
  public viewModel$ = this.select({
    todos$: this.$todos,
    isLoading$: this.$isLoading,
    });
```
And I like to add the debounce property to true on the viewModel only to update it only once.

apps/crud/src/app/services/todo.service.ts
```diff
@@ -0,0 +1,170 @@
...
+      }),
+      concatMap((todoItem) =>
+        this.api.deleteTodo(todoItem.id).pipe(
+          map(() => false),
+          catchError(() => {
+            this.setItemErrorMessage(todoItem.id);
+            return of(true);
+          }),
+          tap((hasError) => !hasError && this.deleteItem(todoItem.id)),
+          finalize(() => this.stopProcessingItem(todoItem.id))
```
tomalaforge, line 59: you have the `tapResponse` operator which combine all of that.

tomalaforge:
Very nice work and very complete solution. great job 💪⭐️

HosseinSalmanian:
@tomalaforge  thank you for reviewing my pr.
your comments was very helpful.

PR #142, opened 2023-09-30
Answer:16 Dependency Injection

edezekiel:

- [x ] Start your PR message with Answer:${challenge_number}

### DIFF
--- a/apps/di/src/app/app.component.ts
+++ b/apps/di/src/app/app.component.ts
@@ -1,6 +1,6 @@
 /* eslint-disable @angular-eslint/directive-selector */
 import { AsyncPipe, NgFor } from '@angular/common';
-import { Component, Directive } from '@angular/core';
+import { Component, Directive, Input, OnInit } from '@angular/core';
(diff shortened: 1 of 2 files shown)


apps/di/src/app/app.component.ts
```diff
@@ -23,10 +23,31 @@ export class ProductDirective {
   }
 }
 
+@Directive({
+  selector: 'tr[currencyCode]',
+  standalone: true,
+  providers: [{ provide: CurrencyService }],
```
tomalaforge, line 29: question: I have never seen this synthax, what is the difference with 
```ts
providers: [CurrencyService]
```

tomalaforge:
👍

apps/di/src/app/app.component.ts
edezekiel: Before I settled on the current solution I was trying some approaches using providers `multi: true` and I forgot to revert this back to `providers: [CurrencyService]`. Fixed! Thanks for reviewing.

PR #146, opened 2023-10-01
Answer:6

flippingtables:
Answer:6

My solution using a Signal based store.


tomalaforge:
Very nit solution. I love the use of computed and effect inside the directive. Didn't think of using it like this. ⭐️ thanks for submitting it.

PR #147, opened 2023-10-01
Answer:12 Optimize Change Detection

tomer953:
I learned this from:
[The danger of using Angular HostListeners](https://blog.simplified.courses/the-danger-of-angular-host-listeners/)


tomalaforge:
Nice read by brecht 👍 and nice solution

PR #150, opened 2023-10-01
Answer:16 Dependency Injection

tomer953:
I know you probably wanted us to creating a directive for each `tr`,  but I used attribute selectors to separate each `tr` to a new component without breaking layout.
Very useful for separating table rows into components.
Learned this from here:
[When to use an attribute selector for Angular components](https://medium.com/javascript-everyday/when-to-use-an-attribute-selector-for-angular-components-7e788ba1bfe7)

### DIFF
--- a/apps/di/src/app/app.component.ts
+++ b/apps/di/src/app/app.component.ts
@@ -1,10 +1,9 @@
 /* eslint-disable @angular-eslint/directive-selector */
-import { AsyncPipe, NgFor } from '@angular/common';
+import { NgFor } from '@angular/common';
 import { Component, Directive } from '@angular/core';
 import { TableModule } from 'primeng/table';
-import { CurrencyPipe } from './currency.pipe';
-import { CurrencyService } from './currency.service';
 import { Product, products } from './product.model';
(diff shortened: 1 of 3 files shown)


tomalaforge:
I am not looking for a specific solution. That's the goal of this repository. My solution was to use a directive and I like your solution too. (your solution is archiving the same as a directive. attribute selector for component is like adding a directive on top of a layout)
 
Your solution is easier to understand for junior I think. Less magical than adding a directive. 
Thanks for submitting it. Give a different solution than mine.

tomer953:
Thanks! 

Angular component **is a** directive (and not vise versa)

So yeah its kind of the same but more intuitive to me.

tomalaforge:
i'll rather say: components are directives with template. 

or directives are components without a template. Sometimes, this way is easier to understand for people starting with the framework. Directives enhances component's behavior

tomer953:
Usually the community refer to "attribute directives" just as "directives", but there are 3 types of built-in directives which components are one of them

source:
https://angular.io/guide/built-in-directives

I just wanted to point this out for people like me, that discovered it not long time ago😀

PR #153, opened 2023-10-02
Answer:1 Projection Challenge

lalit-s-rajput:

- Created City card component
- added templaterefOutLet directive in the card component
- moved control from list and card components to respective original components
- Events added to the list as well as card component
- Removed unnecessary imports

### DIFF
--- a/apps/projection/src/app/ui/card/card.component.html
+++ b/apps/projection/src/app/ui/card/card.component.html
@@ -1,29 +1,27 @@
-<div class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3" [class]="customClass">
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
   <img
     *ngIf="type === CardType.TEACHER"
     src="assets/img/teacher.png"
-    width="200px"
-  />
+    width="200px" />
   <img
     *ngIf="type === CardType.STUDENT"
     src="assets/img/student.webp"
-    width="200px"
-  />
+    width="200px" />
 
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item.firstname ? item.firstname : item.name"
       [id]="item.id"
-      [type]="type"
-    >
+      [type]="type">
     </app-list-item>
   </section>
 
   <button
     class="border border-blue-500 bg-blue-300 p-2 rounded-sm"
-    (click)="addNewItem()"
-  >
+    (click)="addNewItem()">
     Add
   </button>
 </div>
--- a/apps/projection/src/app/ui/card/card.component.html
+++ b/apps/projection/src/app/ui/card/card.component.html
@@ -3,17 +3,11 @@
   [class]="customClass">
   <ng-content select="img"></ng-content>
   <section>
-    <!-- <ng-container [ngTemplateOutlet]="childTemplate"></ng-container> -->
-    <app-list-item *ngFor="let item of list">
-      <ng-template #listTemplate>
-        <div class="border border-grey-300 py-1 px-2 flex justify-between">
-          {{ item.firstname ? item.firstname : item.name }}
-          <button (click)="deleteItem(item.id)">
-            <img class="h-5" src="assets/svg/trash.svg" />
-          </button>
-        </div>
-      </ng-template>
-    </app-list-item>
+    <ng-container
+      *ngFor="let item of list"
+      [ngTemplateOutlet]="listView"
+      [ngTemplateOutletContext]="{ $implicit: item }">
+    </ng-container>
   </section>
 
   <button
--- a/apps/projection/src/app/ui/card/card.component.html
+++ b/apps/projection/src/app/ui/card/card.component.html
@@ -1,17 +1,18 @@
 <div
   class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
   [class]="customClass">
-  <ng-container
-    [ngTemplateOutlet]="
-      templateView ? templateView : defaultView
-    "></ng-container>
+  <ng-content select="img"></ng-content>
   <section>
-    <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname ? item.firstname : item.name"
-      [id]="item.id"
-      [type]="type"
-      (deleteEvent)="deleteItem($event)">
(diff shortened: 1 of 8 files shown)


apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,44 @@
...
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
+  imports: [NgIf, NgFor, ListItemComponent, CommonModule],
+  //providers: [TemplateRef],
 })
 export class CardComponent {
   @Input() list: any[] | null = null;
   @Input() type!: CardType;
   @Input() customClass = '';
+  @Input() templateView: TemplateRef<any>;
```
tomalaforge, line 27: template should almost never be pass as input. 

You can get them using `@ContentChild`

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,25 @@
...
-    width="200px"
-  />
-
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
+  <ng-container
+    [ngTemplateOutlet]="
+      templateView ? templateView : defaultView
+    "></ng-container>
```
tomalaforge, line 7: when you need to project static asset, use `ng-content`, use can use `select` to be more specific on what you want to project

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,44 @@
...
+  @Output() deleteEvent = new EventEmitter<number>();
+  @Output() addEvent = new EventEmitter();
 
   CardType = CardType;
 
   constructor(
     private teacherStore: TeacherStore,
-    private studentStore: StudentStore
+    private studentStore: StudentStore,
+    private cityStore: CityStore
```
tomalaforge, line 36: This component is a shared component, you should not have any logic inside here. 

use `@Output()` to emit event outside a dumb component, and the logic will be handle by the parent

apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -19,17 +17,9 @@ export class ListItemComponent {
   @Input() id!: number;
   @Input() name!: string;
   @Input() type!: CardType;
```
tomalaforge, line 19: same here, all those inputs are not necessary. This component is just a container where you want to project content. (see `ng-content`) and this cimponent will only emit event

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,25 @@
...
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item.firstname ? item.firstname : item.name"
       [id]="item.id"
       [type]="type"
-    >
+      (deleteEvent)="deleteItem($event)">
     </app-list-item>
```
tomalaforge, line 15: look at ng-template-outlet to deal with this part.

tomalaforge:
Thanks for submitting your answer. That's a good first try but you need to read more about `ng-content` and `ngTemplateOutlet`. The first one is quite easy to get, but the second one is more complex. Don't hesitate to ask question on this thread if you need to. 
Good luck

apps/projection/src/app/ui/card/card.component.ts
lalit-s-rajput: done

apps/projection/src/app/ui/card/card.component.html
lalit-s-rajput: done

apps/projection/src/app/ui/card/card.component.ts
lalit-s-rajput: those were unused imports, I missed that and removed them in this commit.

apps/projection/src/app/ui/list-item/list-item.component.ts
lalit-s-rajput: done

apps/projection/src/app/ui/card/card.component.html
lalit-s-rajput: done

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,24 @@
...
-      *ngFor="let item of list"
-      [name]="item.firstname"
-      [id]="item.id"
-      [type]="type"
-    >
+    <!-- <ng-container [ngTemplateOutlet]="childTemplate"></ng-container> -->
+    <app-list-item *ngFor="let item of list">
+      <ng-template #listTemplate>
+        <div class="border border-grey-300 py-1 px-2 flex justify-between">
+          {{ item.firstname ? item.firstname : item.name }}
```
tomalaforge, line 10: You are getting there but not completely. Having conditions inside a shared component should trigger a warning that the pattern use is not the right one. You will end up adding condition each time you add a new component or that the product team wants a new design. 

What is displayed on the row should be done inside the parent element (TeacherCard, CityCard, ... )? To do that, use ngTemplateOutlet with context. 

The delete button should stay inside the ListItemComponent. The way you did it makes the ListItem useless. 

NgTemplateOutlet should be inside this component instead of the ng-template and the template should be inside the parent element. And ListItem wlll be call inside the template in the parent element. Hope this make sense but your code is getting better. Keep up the good work 💪

apps/projection/src/app/ui/card/card.component.html
```diff
@@ -1,29 +1,24 @@
...
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px"
-  />
-
+<div
+  class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3"
+  [class]="customClass">
+  <ng-content select="img"></ng-content>
```
tomalaforge, line 4: very nice 🔥

apps/projection/src/app/ui/card/card.component.html
lalit-s-rajput: I have changed the above as per the review comments and took help from other PR as well as from this video. [Link](https://www.youtube.com/watch?v=_v_MEm9iV-c)

tomalaforge:
That's better, nice job. I hope you learn new things 💪

PR #155, opened 2023-10-02
Change sidebar order

tomer953:
As discussed here: #154 

I implemented sorting by difficulty.

`challenge-number.json` contains counters for each difficulty:
```json
{
  "total": 35,
  "🟢": 12,
  "🟠": 116,
  "🔴": 207
}
```

**Notes**

1.  I chose to use 1, 101, 201 (and not 100, 200) to be consist (since "easy" starting with 1, and not from 0)
2. I first reordered manually all the items by diffulty (first commit), and test it, then moved to the `generator.ts` part

I found **two bugs** in the current implementaion (regardless to the sort order) -
 a. new challanges is created with the wrong index (try to create one now, the filename will be 35-foo, when its need to be 36-foo, since 35 already taken)
  b. even in dry-run mode, two files were created, this is because you used `writeFile` (and `readFile`) from `fs` instead of nx `tree`

I fixed them both.

please review and test


tomalaforge:
Yes that's true about the bugs. I fixed them but didn't push the code 😅 But thanks for fixing it

tomalaforge:
This is really nice and perfect. UX is so much nicer 🔥

Thanks a lot for this PR.

tomalaforge (approved):

tomalaforge:
@all-contributors add @tomer953 for doc,code

PR #159, opened 2023-10-03
Answer:8

HosseinSalmanian:
A pipe named heavyComputationPipe was added
in order to avoid using function call in template.


tomalaforge:
That's perfect, nice job 👍

PR #160, opened 2023-10-03
Answer:21

HosseinSalmanian:
I used router link and router link in order to prevent refreshing the page while navigation to another page and used fragment and enabled anchor scrolling for smooth scrolling.


tomalaforge:
Thanks for your answer 🔥

PR #161, opened 2023-10-03
Answer:22

HosseinSalmanian:
test component refactored using router input instead activated route


tomalaforge:
That's perfect and so much nicer, isn't it ? 😍

PR #162, opened 2023-10-03
Answer:29 testing todos list

jdegand:
Done can only be updated once. However, you can change the assignee after the todo is already done. Unintended bug?

I used Testing Library. 

It is tough finding great resources to help complete this challenge.  This [article](https://timdeschryver.dev/blog/testing-an-ngrx-project#unit-tests) from Tim Deschryver is must read for anyone struggling writing these tests.  

I also created a separate [repo](https://github.com/jdegand/testing-todos-list/) for this challenge with a lot of resources I referenced to complete this.  

I left a lot of comments in the code so people can see my thought process and some of the tradeoffs you have to make when testing.  

I have to point out that in your solution you have awaited screen queries.  Although it is tough to find a definitive answer, I believe this is incorrect.  You should use waitFor.  At one point, I also used  fixture.detectChanges to get around async issues.   

I didn't include userEvent in my setup function.  That is something that could be quickly fixed.

I added Backend Service tests in lieu of mostly redundant store tests. 

I used an alternative way to test the route navigation but it was slower than your solution.

### DIFF
--- a/apps/testing-todos-list/src/app/list/ticket.store.spec.ts
+++ b/apps/testing-todos-list/src/app/list/ticket.store.spec.ts
@@ -1,11 +1,47 @@
+import { render } from '@testing-library/angular';
+import { createMockWithValues } from '@testing-library/angular/jest-utils';
+import { of } from 'rxjs';
+import { APP_ROUTES } from '../app.route';
+import { BackendService } from '../backend.service';
+import { ListComponent } from './list.component';
+import { TicketStore } from './ticket.store';
+import { provideComponentStore } from '@ngrx/component-store';
+import { TestBed } from '@angular/core/testing';
+
+const USERS = [
+  { id: 1, name: 'titi' },
+  { id: 2, name: 'george' },
+];
+
+const TICKETS = [
+  {
+    id: 0,
+    description: 'Install a monitor arm',
+    assigneeId: 1,
+    completed: false,
+  },
+  {
+    id: 1,
+    description: 'Coucou',
+    assigneeId: 1,
+    completed: false,
+  },
+];
+
+// TicketStore coverage is already good from list tests.  (I did the list tests before the store tests)
+// Will these tests improve code coverage? No.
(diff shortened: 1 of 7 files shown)


apps/testing-todos-list/src/app/list/ticket.store.spec.ts
```diff
@@ -16,7 +52,10 @@ describe('TicketStore', () => {
 
     describe('Given users api returns failure response', () => {
       it('Then tickets should not have any assignee', () => {
-        //
+        // I think this is the test that will help with line 50
+        // Not really sure how I would test this as Thomas' solution has a more involved setup function
+        // mockBackendService.users.and.returnValue(of(throwError()=> new Error("Blah blah blah"))) or something close
+        // then check tickets are the same as TICKETS
```
tomalaforge, line 58: did you push everything? There is no test here!!!

tomalaforge:
Thanks for your answer. Did you push everything ?

Just want to understand this remark:
"I have to point out that in your solution you have awaited screen queries. Although it is tough to find a definitive answer, I believe this is incorrect. You should use waitFor. At one point, I also used fixture.detectChanges to get around async issues."

What do you mean awaiting queries is wrong. `findBy` is implementing `waitFor` under the hood. You should almost never use `waitFor` directly. It's always better to wait for element to appear or disappear.

To debug or find the correct selector, you should read this [page](https://testing-library.com/docs/dom-testing-library/api-debugging/#screenlogtestingplaygroundurl)

apps/testing-todos-list/src/app/list/ticket.store.spec.ts
jdegand: Pseudo code.  Didn't write a test.

jdegand:
The code looked awkward.  I guess I got getBy and findBy confused.  I did not consult the docs for clarity.  I noted it and moved on.

PR #165, opened 2023-10-04
docs: update start command

DeveshChau:
updated the start command from 
`npx nx serve ngfor-optimize` to
`npx nx serve performance-ngfor-optimize`

### DIFF
--- a/libs/cli/src/generators/challenge/generator.ts
+++ b/libs/cli/src/generators/challenge/generator.ts
@@ -41,7 +41,10 @@ function findPreviousChallengeFilePath(tree, path, number) {
 }
 
 export async function challengeGenerator(tree: Tree, options: Schema) {
-  const { appDirectory } = getProjectDir(options.name, options.directory);
+  const { appProjectName, appDirectory } = getProjectDir(
+    options.name,
+    options.directory
+  );
 
   const difficulty = options.challengeDifficulty;
 
@@ -73,7 +76,7 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
 
   generateFiles(tree, join(__dirname, 'files', 'readme'), appDirectory, {
     tmpl: '',
-    projectName: names(options.name).name,
+    projectName: appProjectName,
     title: options.title,
     challengeNumber,
     docRepository: options.docRepository,
@@ -85,7 +88,7 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
     `./docs/src/content/docs/challenges/${options.docRepository}`,
     {
       tmpl: '',
-      projectName: names(options.name).name,
+      projectName: appProjectName,
       title: options.title,
       challengeNumber,
       difficulty,
--- a/libs/cli/src/generators/challenge/generator.ts
+++ b/libs/cli/src/generators/challenge/generator.ts
@@ -76,7 +76,8 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
 
   generateFiles(tree, join(__dirname, 'files', 'readme'), appDirectory, {
     tmpl: '',
-    projectName: appProjectName,
+    projectName: names(options.name).name,
+    appProjectName: appProjectName,
     title: options.title,
     challengeNumber,
     docRepository: options.docRepository,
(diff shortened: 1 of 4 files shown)


DeveshChau:
Yes sure, I will be happy to correct it.
Please guide me on this.

tomalaforge:
nice 💪

You can create generators inside Nx to script redundant command. All generators are located inside a Nx plugin which is called `cli` inside *angular challenge* that you can find in the lib folder.

The generator that interest us is the one called `challenge` (path: `libs/cli/src/generators/challenge/generator.ts`). This is the script file.

To run the script, the easier way is to install the Nx console plugin inside vscode. Once done, click on Nx generate > @angular-challenge:cli - challenges

Fill all fields and you'll see a dryrun running inside your terminal. 

Now that you know how this is working. you can try to debug the file. There is also 2 challenges on Nx generator inside this repo if you want to practise and learn more about them.

If you have any questions, you can ask them here or DM me on twitter for an easier chat.

DeveshChau:
Thanks for the details. I will surely try this one.

DeveshChau:
I used appProjectName to update projectName.

libs/cli/src/generators/challenge/generator.ts
```diff
@@ -41,7 +41,10 @@ function findPreviousChallengeFilePath(tree, path, number) {
 }
 
 export async function challengeGenerator(tree: Tree, options: Schema) {
-  const { appDirectory } = getProjectDir(options.name, options.directory);
+  const { appProjectName, appDirectory } = getProjectDir(
+    options.name,
+    options.directory
+  );
```
tomalaforge, line 47: that's awesome. I knew that Nx has this name already figured out. 👍

libs/cli/src/generators/challenge/generator.ts
```diff
@@ -73,7 +76,7 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
 
   generateFiles(tree, join(__dirname, 'files', 'readme'), appDirectory, {
     tmpl: '',
-    projectName: names(options.name).name,
+    projectName: appProjectName,
```
tomalaforge, line 79: be careful. You need to create a new property. We still need projectName inside our file. 

look inside: `libs/cli/src/generators/challenge/files/readme/README.md__tmpl__` to understand how those properties are used

libs/cli/src/generators/challenge/generator.ts
```diff
@@ -85,7 +88,7 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
     `./docs/src/content/docs/challenges/${options.docRepository}`,
     {
       tmpl: '',
-      projectName: names(options.name).name,
+      projectName: appProjectName,
```
tomalaforge, line 91: to be more consistent, you should rename `projectName` into `appProjectName` (dont forget to do it inside the doc file.

libs/cli/src/generators/challenge/generator.ts
DeveshChau: I thought of creating a new property, but I observe that `projectName` is being used in 2 files where value of `projectName` should be updated.
`libs/cli/src/generators/challenge/files/docs/__challengeNumber__-__projectName__.md__tmpl__`
and 
`libs/cli/src/generators/challenge/files/readme/README.md__tmpl__`.

doc file is generated with `projectName`is like **`37-performance-demo.md`**

That is why I thought to go with this approach.

Let me know if I should create new property `appProjectName`,
 or rename `projectName` to  `appProjectName` and use it in
 `README.md__tmpl__`  and `__challengeNumber__-__projectName__.md__tmpl__`

tomalaforge: Yes that's true. But I prefer to name the new file `37-demo.md` instead of the long name with the directory prefix. 

So having two separate properties would be better. 

Thanks for your work and time.

DeveshChau: Done.
Thank you for the opportunity.

libs/cli/src/generators/challenge/generator.ts
```diff
@@ -86,6 +90,7 @@ export async function challengeGenerator(tree: Tree, options: Schema) {
     {
       tmpl: '',
       projectName: names(options.name).name,
+      appProjectName: appProjectName,
```
tomalaforge, line 93: nit/non blocking: you don't have to repeat the property 😅

tomalaforge (approved):

tomalaforge:
@all-contributors add @DeveshChau for code,bug

PR #167, opened 2023-10-04
feat(Answer:1): solution projection

tomalaforge:


apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,44 @@
...
+    </button>
+  `,
+  standalone: true,
+  imports: [NgIf, NgFor, NgTemplateOutlet],
+  host: {
+    class: 'border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3',
+  },
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class CardComponent<T> {
```
luchillo17, line 38: May I ask how you tell `T` it has an `id` prop? using the `@for` loop now needs a `track` param, so this gives me an error:
image

luchillo17: Normally I would tell `T` it expects an `id` prop like this, but then it breaks in the city card component:
```ts
export class CardComponent<T extends {id: string}> {
```
image

luchillo17: Nevermind found my issue, id is a number, so the types were not overlapping...

apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,23 @@
...
+} from '@angular/core';
 
 @Component({
   selector: 'app-list-item',
   template: `
     <div class="border border-grey-300 py-1 px-2 flex justify-between">
-      {{ name }}
-      <button (click)="delete(id)">
+      <ng-content></ng-content>
+      <button (click)="delete.next()">
```
MansurIsakov, line 13: Hey, I have a question, why did you use delete.next() instead of delete.emit(), Ive researched this and nextCall() seems to be deprecated

https://github.com/angular/angular/blob/b5b6ece65a96f5b8f134ad4899b56bf84afe3ba0/modules/angular2/src/facade/async.dart#L49

PR #168, opened 2023-10-04
Answer 34

BillyOkoth:
1. Create smaller components with on push change detection
2. Implementation of Application ref to log  change detection


apps/performance/default-onpush/src/app/search.component.ts
```diff
@@ -0,0 +1,37 @@
...
+@Component({
+  selector: 'app-search',
+  standalone: true,
+  imports: [
+    MatFormFieldModule,
+    MatInputModule,
+    CDFlashingDirective,
+    FormsModule,
+  ],
+  providers: [CDFlashingDirective],
```
tomalaforge, line 16: question: why did you put it inside the providers array ?

tomalaforge:
Nice job 💪 This is working fine. I'm just intrigue by the `enableDebugTools` inside the `main.ts`. Why do you need that ?

PR #169, opened 2023-10-04
Answer:12

BillyOkoth:
1. Running Ngzone outside angular to optimize change detection

### DIFF
--- a/apps/scroll-cd/src/app/app.component.ts
+++ b/apps/scroll-cd/src/app/app.component.ts
@@ -1,5 +1,5 @@
 import { AsyncPipe, NgIf } from '@angular/common';
-import { Component, HostListener } from '@angular/core';
+import { Component, HostListener, NgZone, inject } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
 
 @Component({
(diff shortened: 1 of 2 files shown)


apps/scroll-cd/src/main.ts
```diff
@@ -1,4 +1,16 @@
...
+} from '@angular/platform-browser';
 import { AppComponent } from './app/app.component';
+import { ApplicationRef } from '@angular/core';
 
-bootstrapApplication(AppComponent).catch((err) => console.error(err));
+bootstrapApplication(AppComponent)
+  .then((module) => {
+    const appRef = module.injector.get(ApplicationRef);
+    const appComponent = appRef.components[0];
+    enableDebugTools(appComponent);
```
tomalaforge, line 12: question: what do you need to add this ? What browser do you use ?

apps/scroll-cd/src/app/app.component.ts
```diff
@@ -33,14 +33,21 @@ import { BehaviorSubject } from 'rxjs';
...
-
+  ngZone: NgZone = inject(NgZone);
   private displayButtonSubject = new BehaviorSubject<boolean>(false);
   displayButton$ = this.displayButtonSubject.asObservable();
 
+  ngOnInit() {
+    this.ngZone.runOutsideAngular(() => {
+      window.addEventListener('scroll', this.onScroll.bind(this));
+    });
+  }
```
tomalaforge, line 44: This is not working. How did you test it ? Cd is still triggered at each CD cycle

PR #171, opened 2023-10-04
Answer:32 bug cd memo

jdegand:
This is sort of an alternative solution.  

I used this [article](https://itnext.io/its-ok-to-use-function-calls-in-angular-templates-ffdd12b0789e) for the memo function and the idea.  

I also thought ngOnChanges could be a possible solution or am I wrong ?


tomalaforge:
The two solutions works but they are just hiding the problem. 

Did you read my article on this challenge [here](https://medium.com/ngconf/function-calls-inside-template-are-dangerous-15f9822a6629) ? 

The issue comes from the fact that the function return a new instance of the array at each CD cycle. TrackBy is a good addon, but the main issue should be fixed first. 👍

PR #173, opened 2023-10-05
Answer:14

mohammadghaderi:

### DIFF
--- a/apps/rxjs-race-condition/src/app/app.component.ts
+++ b/apps/rxjs-race-condition/src/app/app.component.ts
@@ -1,37 +1,33 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  inject,
-  OnInit,
-} from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
-import { take } from 'rxjs';
+import { Observable, take } from 'rxjs';
 import { TopicModalComponent } from './topic-dialog.component';
 import { TopicService, TopicType } from './topic.service';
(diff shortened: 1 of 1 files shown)


apps/rxjs-race-condition/src/app/app.component.ts
```diff
@@ -1,37 +1,33 @@
...
 })
-export class AppComponent implements OnInit {
+export class AppComponent {
   title = 'rxjs-race-condition';
   dialog = inject(MatDialog);
   topicService = inject(TopicService);
-  topics: TopicType[] = [];
+  topics$: Observable<TopicType[]> = this.topicService
+    .fakeGetHttpTopic()
+    .pipe(take(1));
```
tomalaforge, line 25: question: why did you use a take 1 here ?

apps/rxjs-race-condition/src/app/app.component.ts
```diff
@@ -1,37 +1,33 @@
...
-      .pipe(take(1))
-      .subscribe((topics) => (this.topics = topics));
-  }
-
-  openTopicModal() {
+  openTopicModal(topics: TopicType[]) {
     this.dialog.open(TopicModalComponent, {
       data: {
-        topics: this.topics,
+        topics: topics,
```
tomalaforge, line 30: nit: don't repeat when the name is identical

apps/rxjs-race-condition/src/app/app.component.ts
mohammadghaderi: I forgot to delete the pipe as we don't need it. just pushed new changes

PR #175, opened 2023-10-05
fix: missing import crud setup file

jdegand:
If you try to create a spec file and add describe, you get a warning about jest types missing.   Setup file is missing the import.


tomalaforge:
that's write, at that time, the config for testing lib was not added. I'll add it, there is some other files to change

tomalaforge:
Closing the issue but thanks for reporting it

tomalaforge:
Closing the issue but thanks for reporting it

PR #179, opened 2023-10-06
Answer:1

reed756:
- Amended Card component and List item component. Also added in CityCard component.


apps/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,21 @@
...
-import { StudentStore } from '../../data-access/student.store';
-import { TeacherStore } from '../../data-access/teacher.store';
-import { CardType } from '../../model/card.model';
-
+import { Component, EventEmitter, Input, Output } from '@angular/core';
 @Component({
   selector: 'app-list-item',
   template: `
     <div class="border border-grey-300 py-1 px-2 flex justify-between">
       {{ name }}
```
tomalaforge, line 6: to make it more shareable, leave the information you want to put inside here to the parent

apps/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,31 @@
...
 })
 export class CardComponent {
   @Input() list: any[] | null = null;
-  @Input() type!: CardType;
   @Input() customClass = '';
+  @Output() add = new EventEmitter<void>();
+  @ContentChild('rowRef', { read: TemplateRef })
+  rowTemplate!: TemplateRef<{ $implicit: any }>;
 
   CardType = CardType;
```
tomalaforge, line 26: nit:to delte

apps/projection/src/styles.scss
```diff
@@ -1,3 +1,15 @@
...
+  .bg-light-red {
+    background-color: rgba(250, 0, 0, 0.1);
+  }
+  .bg-light-green {
+    background-color: rgba(0, 250, 0, 0.1);
+  }
+  .bg-light-yellow {
+    background-color: rgba(230, 234, 12, 0.301);
+  }
+}
```
tomalaforge, line 15: When the app is small, that's ok. but It will create a bloated file. and the style are far from where there are used. prefer to keep them as close as their component. 

only global style should come here

tomalaforge:
Nice job, I added some small comments but otherwise, this is great work

apps/projection/src/app/ui/list-item/list-item.component.ts
reed756: ok, so put an ng-content in the list item where {{name}} is and then put the name in the specific card component?

apps/projection/src/app/ui/card/card.component.ts
reed756: I've deleted the card type now.

reed756:
Thanks for your review and for the pointers you've given :)

PR #180, opened 2023-10-06
Answer:36 solution to ngfor-optimize

tomer953:
One day I'll tell the kids: "look how ugly it was to add trackBy function"
+ Its almost impossible to remember the right syntax without snippets for me.

Soon, angular v17 will introduce a more straight forward way to trackBy custom property directly in the template!
👏

Until then, you can use a utility that does the same:
https://ngxtension.netlify.app/utilities/directives/trackby-id-prop/

### DIFF
--- a/apps/performance/ngfor-optimize/src/app/person-list.component.ts
+++ b/apps/performance/ngfor-optimize/src/app/person-list.component.ts
@@ -9,7 +9,7 @@ import { Person } from './person.model';
   imports: [CommonModule],
   template: `
     <div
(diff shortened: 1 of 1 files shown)


tomalaforge:
don't need to use ngxtension. The utility is already inside angular challenges. The guy who implemented it, copy/paste from my article: https://medium.com/ngconf/make-trackby-easy-to-use-a3dd5f1f733b without changing a word 😅

And yes, the new control flow will make that part super easy 👍

PR #182, opened 2023-10-06
Answer:23 solution to harness testing

tomer953:
I compared with your sol, seems fine except that I used the `component.max` `component.value` and similar in some cases rather then query the material harness api, wdyt ?

### DIFF
--- a/apps/testing-harness/src/app/child.component.spec.ts
+++ b/apps/testing-harness/src/app/child.component.spec.ts
@@ -1,38 +1,84 @@
 import { render } from '@testing-library/angular';
+import { TestbedHarnessEnvironment } from '@angular/cdk/testing/testbed';
+import { MatButtonHarness } from '@angular/material/button/testing';
+import { MatCheckboxHarness } from '@angular/material/checkbox/testing';
+import { MatInputHarness } from '@angular/material/input/testing';
+import { MatSliderHarness } from '@angular/material/slider/testing';
+import { HarnessLoader } from '@angular/cdk/testing';
 import { ChildComponent } from './child.component';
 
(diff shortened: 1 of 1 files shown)


apps/testing-harness/src/app/child.component.spec.ts
```diff
@@ -1,38 +1,84 @@
...
+
   describe('When init', () => {
     test('Then show 1 slider, 3 checkboxes, 4 inputs, 2 buttons', async () => {
-      await render(ChildComponent);
+      const [inputs, checkboxes, sliders, buttons] = await Promise.all([
+        loader.getAllHarnesses(MatInputHarness),
+        loader.getAllHarnesses(MatCheckboxHarness),
+        loader.getAllHarnesses(MatSliderHarness),
+        loader.getAllHarnesses(MatButtonHarness),
+      ]);
```
tomalaforge, line 26: really nice, haven't thought of using `Promise.all`

apps/testing-harness/src/app/child.component.spec.ts
```diff
@@ -1,38 +1,84 @@
...
+      const value = await endThumb.getValue();
+      expect(value).toBe(0);
     });
   });
 
   describe('Given maxValue set to 109', () => {
     test('Then slider max value is 109', async () => {
-      await render(ChildComponent);
+      const slider = await loader.getHarness(MatSliderHarness);
+      component.max = 109;
```
tomalaforge, line 45: The thing I don't like is that you use implementation detail. I always avoid using it and I create my test from the user point of view and I cannot use anything else that the user is not seeing. This way, if the implementation change, my test should always pass.

PR #185, opened 2023-10-09
Answer:5 Crud with tests

jdegand:
This is slightly reworked from [my repo](https://github.com/jdegand/crud-app-local-store) that used karma and jasmine.

### DIFF
--- a/apps/crud/src/app/app.component.ts
+++ b/apps/crud/src/app/app.component.ts
@@ -1,51 +1,34 @@
-import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
-import { randText } from '@ngneat/falso';
+import { CommonModule, NgIf } from '@angular/common';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
+import { AppStore } from './app.store';
+import { provideComponentStore } from '@ngrx/component-store';
+import { LoadingSpinnerComponent } from './components/loading-spinner/loading-spinner.component';
+import { TodoItemComponent } from './components/todo-item/todo-item.component';
+
+// can render vm without ngrxLet
+// is ngrxLet really necessary?
+// Todos are not gonna be zero / have to delete 200 todos to get to zero
 
 @Component({
   standalone: true,
-  imports: [CommonModule],
+  imports: [CommonModule, LoadingSpinnerComponent, NgIf, TodoItemComponent],
+  providers: [provideComponentStore(AppStore)],
   selector: 'app-root',
+  changeDetection: ChangeDetectionStrategy.OnPush,
   template: `
-    <div *ngFor="let todo of todos">
-      {{ todo.title }}
-      <button (click)="update(todo)">Update</button>
-    </div>
+    <ng-container *ngIf="vm$ | async as vm">
+      <div *ngIf="vm.callState.toLowerCase() === 'loading'">
+        <app-loading-spinner></app-loading-spinner>
+      </div>
+      <app-todo-item
+        *ngFor="let todo of vm.todos"
+        [todo]="todo"></app-todo-item>
+    </ng-container>
   `,
   styles: [],
 })
-export class AppComponent implements OnInit {
-  todos!: any[];
+export class AppComponent {
+  private appStore = inject(AppStore);
 
-  constructor(private http: HttpClient) {}
-
-  ngOnInit(): void {
-    this.http
-      .get<any[]>('https://jsonplaceholder.typicode.com/todos')
-      .subscribe((todos) => {
-        this.todos = todos;
-      });
-  }
-
-  update(todo: any) {
-    this.http
-      .put<any>(
-        `https://jsonplaceholder.typicode.com/todos/${todo.id}`,
-        JSON.stringify({
-          todo: todo.id,
-          title: randText(),
-          body: todo.body,
-          userId: todo.userId,
-        }),
-        {
-          headers: {
-            'Content-type': 'application/json; charset=UTF-8',
-          },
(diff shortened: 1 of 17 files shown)


apps/crud/src/app/app.store.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// maybe better to add {providedIn: root} to Injectable
+// is it really necessary to create another store just for a presentational component?
+// pass update / delete methods to TodoItemComponent?
+@Injectable()
+export class AppStore
+  extends ComponentStore<AppState>
+  implements OnStateInit, OnStoreInit
+{
+  // making todo$ and callState$ -> private -> makes testing more difficult ?
```
tomalaforge, line 27: To test private method you can always do:
`(todoService as any)['myPrivateMethod']` 
but this is not very nice indeed

apps/crud/src/app/app.store.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  readonly vm$ = this.select(
+    {
+      todos: this.todos$,
+      callState: this.callState$,
+    },
+    { debounce: true }
+  );
+
+  // this is outdated ngrx syntax - don't need to use a trigger$ anymore
```
tomalaforge, line 51: So why did you use it 😅 🤣

apps/crud/src/app/components/loading-spinner/loading-spinner.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="lds-ring">
+    <div></div>
+    <div></div>
+    <div></div>
+    <div></div>
+</div>
```
tomalaforge, line 6: first time I see this to create a loader 💪

apps/crud/src/app/components/todo-item/todo-item.component.html
```diff
@@ -0,0 +1,15 @@
+<div *ngIf="vm$ | async  as vm">
```
tomalaforge, line 1: nit: you don't need a div here, can save some nesting DOM element

apps/crud/src/app/components/todo-item/todo-item.component.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    fireEvent.click(button);
+
+    expect(mockTodoService.updateTodo).toHaveBeenCalledTimes(1);
+  });
+
+  test('delete', async () => {
+    const { mockTodoService, fixture } = await setup();
+
+    fixture.detectChanges();
```
tomalaforge, line 34: if you are using TestingLibrary, you should not be using internal implementation. Only what the user sees on the screen can be queried

Why did you need to use this ? 

If you need to wait for the button to appear, you can use, `findByRole`

apps/crud/src/app/components/todo-item/todo-item.component.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    const button = screen.getAllByRole('button')[1];
```
tomalaforge, line 36: getting all buttons is dangerous, because if you add another buttons, this can break the test and create a false positive. 

Prefer `screen.getByRole('button', {name: 'xxx'})`

apps/crud/src/app/components/todo-item/todo-item.component.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    fireEvent.click(button);
```
tomalaforge, line 38: you also have `userEvent`. It reproduices the user interaction better. For instance, if a button is not visible, userEvent will fail wheareas fireEvent will fire the event.

apps/crud/src/app/todoItem.store.ts
```diff
@@ -0,0 +1,68 @@
...
+  callState?: string;
+}
+
+@Injectable()
+export class TodoItemStore extends ComponentStore<TodoItemState> {
+  private todoService = inject(TodoService);
+  private appStore = inject(AppStore);
+
+  // replace with ngrxOnInit ? or constructor with super?
+  // probably should use ngrxOnInit but I'll leave this
```
tomalaforge, line 19: ngOnInit will never get call, you are inside an Injectable. Only ngOnDestroy is called. 

Use a contructor or ngrxOnInit. => that's why your patchState is not orking. Your state is never initialized

apps/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,43 @@
...
+import { provideComponentStore } from '@ngrx/component-store';
+//import { LoadingSpinnerComponent } from './components/loading-spinner/loading-spinner.component';
+import { TodoItemComponent } from './components/todo-item/todo-item.component';
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+
+// originally used loading spinner component from loading.io - swapped for mat spinner
+// can freely swap them
+
+// can render vm without ngrxLet
+// is ngrxLet really necessary?
```
tomalaforge, line 13: no ngIf and ngrxLet are pretty similar.

apps/crud/src/app/app.store.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      const t = await lastValueFrom(store.todos$.pipe(take(1)));
+
+      expect(t.length).toEqual(todos.length);
+    });
+  });
+
+  // can't do success and failure
+  // setup function has only one mocked returned value
+  // and you can't overwrite it inside a test
```
tomalaforge, line 38: you can pass argument to your setup function. That's why it's better than a beforeEch

tomalaforge:
really nice work. Well coded and lots of test. Great job 👍

PR #186, opened 2023-10-10
Angular: 5

DeveshChau:
Tried to implement Step 1 from challenge 5, 
created interface for Todo, Todo service to make http call and used behavior subject. Avoided mutation.

### DIFF
--- a/apps/crud/src/app/loader.interceptor.ts
+++ b/apps/crud/src/app/loader.interceptor.ts
@@ -1,0 +1,31 @@
+import {
+  HttpErrorResponse,
+  HttpEvent,
+  HttpHandler,
+  HttpInterceptor,
+  HttpRequest,
+} from '@angular/common/http';
+import { Observable, catchError, finalize, of, throwError } from 'rxjs';
+import { LoaderService } from './loader.service';
+import { Injectable } from '@angular/core';
+
+@Injectable({ providedIn: 'root' })
+export class LoaderInterceptor implements HttpInterceptor {
+  constructor(private loaderService: LoaderService) {}
+  intercept(
+    req: HttpRequest<any>,
+    next: HttpHandler
+  ): Observable<HttpEvent<any>> {
+    this.loaderService.showLoader();
+    return next.handle(req).pipe(
+      finalize(() => {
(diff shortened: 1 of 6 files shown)


apps/crud/src/app/loader.interceptor.ts
```diff
@@ -0,0 +1,31 @@
...
+        this.loaderService.hideLoader();
+      }),
+      catchError((error) => {
+        this.loaderService.hideLoader();
+        console.log(error);
+        return of();
+      })
+    );
+  }
+}
```
tomalaforge, line 31: question: Did you implement it inside an entreprise application ? Personnaly, I always avoided it because it make the behavior too global. But that's a nice implementation 👍

tomalaforge:
nice work and nice implementation of the interceptors. 👍

apps/crud/src/app/loader.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, Observable, catchError, throwError } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class LoaderService {
+  loaderSubject = new BehaviorSubject<boolean>(false);
```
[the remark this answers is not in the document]
DeveshChau: Got it, I will make it private.

apps/crud/src/app/loader.interceptor.ts
DeveshChau: Previously I used this approach in my side project.
What could be a better approach other than state management?

tomalaforge: I mean, this is nice approach but all http request will trigger the loader, sometimes, you want it more local to a component to trigger a loader only on a row for instance, you don't want a global loader. With your approach, you don't have that flexibility.

DeveshChau: Ahh got it, as I am progressing with each step from the challenge. Step 2 suggests adding a global loader. Step 4 is expecting to implement state management and localised loader error. I will implement that in the next commit.

PR #189, opened 2023-10-12
Answer:37 - Biglist

jdegand:
Not really sure how many items you wanted displayed but I took the hint of 20 from the readme.  

Used minBufferPx / maxBufferPx to improve performance even more.  
 
Not much info about using trackBy with cdk-virtual-scroll-viewport.  The docs mentioned you just do it like you normally would.

### DIFF
--- a/apps/performance/ngfor-biglist/project.json
+++ b/apps/performance/ngfor-biglist/project.json
@@ -20,7 +20,8 @@
           "apps/performance/ngfor-biglist/src/assets"
         ],
         "styles": ["apps/performance/ngfor-biglist/src/styles.scss"],
-        "scripts": []
+        "scripts": [],
+        "allowedCommonJsDependencies": ["seedrandom"]
       },
       "configurations": {
         "production": {
--- a/apps/performance/ngfor-biglist/project.json
+++ b/apps/performance/ngfor-biglist/project.json
@@ -19,7 +19,10 @@
           "apps/performance/ngfor-biglist/src/favicon.ico",
           "apps/performance/ngfor-biglist/src/assets"
         ],
-        "styles": ["apps/performance/ngfor-biglist/src/styles.scss"],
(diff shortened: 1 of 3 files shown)


apps/performance/ngfor-biglist/src/app/person-list.component.ts
```diff
@@ -27,4 +31,8 @@ import { Person } from './person.model';
 })
 export class PersonListComponent {
   @Input() persons: Person[] = [];
+
+  trackPerson(index: number, person: Person) {
+    return person ? person.email : undefined;
```
tomalaforge, line 36: question: why did you do this ternary statement? Can person be undefined?

apps/performance/ngfor-biglist/project.json
```diff
@@ -19,8 +19,12 @@
...
-        "scripts": []
+        "styles": [
+          "./node_modules/@angular/material/prebuilt-themes/indigo-pink.css",
+          "apps/performance/ngfor-biglist/src/styles.scss"
+        ],
+        "scripts": [],
+        "allowedCommonJsDependencies": ["seedrandom"]
       },
       "configurations": {
         "production": {
```
tomalaforge, line 30: You can rebase to update this file 👍 Thanks for your commit

apps/performance/ngfor-biglist/src/app/person-list.component.ts
jdegand: Copy & Pasted from another project.  I forgot to add the index parameter so I just grabbed a similar trackBy I had recently used.

tomalaforge: You can also use my trackBy directive that simplify the work. `libs/shared/directives/src/lib/track-by.directive.ts`

PR #193, opened 2023-10-13
feat: challenge 38 - rxjs catch error

DeveshChau:
A challenge to learn catchError Rxjs Operator.


jdegand:
- Good setup to the challenge
- Little easy but still important topic
- Adding test requirement could make harder
- Have to click the button for submission to happen on the input - enter doesn't work
- Using button refs -> reason for poor accessibility ?
- Didn't use tailwind for the styling - could be converted - I'm not a big tailwind user 
- Better to cap the width of the returned json from the api request to improve styling
- app.component.ts line 12 spelling error of possible
- 38-rxjs-catch-error.md line 22 need to remove hashtags so paragraph isn't bolded as a heading

DeveshChau:
@jdegand 
Thanks for the review.
Using button refs -> reason for poor accessibility? 
Answer: I want to create subscription once at the time on initiation, hence using button ref. I tried attaching click event of button but on every click new subscription is getting added. Which is making multiple Api calls every time I click on button. Do suggest efficient way to achieve this. I am keen to learn.

Didn't use tailwind for the styling - could be converted - I'm not a big tailwind user.
Reply: Even I haven't used taliwind till now, I guess at the time of generate new challenge tailwind is configured by default.

I cap the width, spelling updated to possible and removed hashtag.

tomalaforge:
😍😍😍😍😍😍😍

Thanks @DeveshChau  for submitting this first challenge contribution. I'll look at it more closely and try to resolve it when I have more time. 👍🚀

Thanks @jdegand for reviewing.

tomalaforge:
I'm sorry but I didn't really get what I have to do. First by reading the doc, I though there were a bug that I need to fix. 

If i guess right, the code is imperative right now, and we need to refacto it to declarative using catchError?

First case: create a buggy application that complete the subscription that we need to correct. In second case, I think you should make the documentation clearer of your intention. 

About tailwind: it's configured by default but you don't have to use it. Maybe you should put your css inside an external file, since it's irrelevant to the challenge. 

a11y is not the theme of the challenge, so that's ok for me if you cannot use the Enter button to submit your form. But I don't get why you used `ViewChild`, you should use a form and use the button click handler. 😅 Your challenge is not on refactoring to use form, so it's better to do it the right way from the beginning. Use a subject to submit your form and trigger your observable. 

Thanks for your challenge. Even it's quite simple, it's a very important topic and many junior dev struggle with observable, so it's a good one.

jdegand:
If you input anything but the listed words above the input, you are prevented from resubmitting the form since it errors out.   You would have to refresh the page to get another chance.  

If you add catchError in the right place with right callback, you can resubmit the form even if your input causes an api failure.

tomalaforge:
Ok I get it, now, but the title is very misleading. 

I think the challenge should only be to refactor to proper Angular using reactiveForm or templateForm and a declarative style, which should correct the catchError. 

I really like that challenge because error handling is a difficult topic often badly managed. We just need to rework the instruction IMO. What do you think ?

jdegand:
You could piggyback off this challenge to create another challenge about refactoring.  But I think this challenge is designed to highlight catchError alone.  Hopefully, other people will give feedback and you could see their suggestions.

tomalaforge:
Ok I get it. But if we are only focusing on catchError, the rest should be proper code. The goal of this repository is to show good practices. 

If we want to focus on catchError and a good refactoring, I agree with that code but it needs to be better explain inside the instruction. 

In the way it's done actually, it's hard to know what to do.

tomalaforge:
about your last thought: what do you mean ? A tag on the current pull request to get more feedback ? Or something else ?

tomalaforge:
One more thing: you will need to rebase and change a few thing on your doc page. Add some meta data and delete some code. Copy one of the other challenge documentation. It should be quite straight forward.

DeveshChau:
My aim is to achieve correct positioning of catch error to avoid completion of subscription, 
something like `fromEvent().pipe(map(), cancatMap(ajax().pipe( catchError()))`.
Using forms, I am not able to generate scenario where user will not be able to send new request.

tomalaforge:
If you want to use formControl, you can use a subject like in the following snippet

```ts
@Component({
  standalone: true,
  imports: [CommonModule, FormsModule],
  selector: 'app-root',
  template: `
    <div class="form-container">
      <span
        >possible values: posts, comments, albums, photos, todos, users</span
      >
    </div>
    <form class="form-container" (ngSubmit)="submit$$.next()">
      <input
        type="text"
        placeholder="Enter text"
        [(ngModel)]="input"
        name="action" />
      <button>Fetch</button>
    </form>
    <div class="response">
      {{ response | json }}
    </div>
  `,
})
  submit$$ = new Subject<void>();

  input = '';

  response: unknown;
  constructor(private http: HttpClient) {}

  ngOnInit() {
    this.submit$$
      .pipe(
        map(() => this.input),
        concatMap((value) =>
          this.http.get(`https://jsonplaceholder.typicode.com/${value}/1`)
        )
      )
      .subscribe({
        next: (value) => {
          console.log(value);
          this.response = value;
        },
        error: (error) => {
          console.log(error);
          this.response = error;
        },
        complete: () => console.log('done'),
      });
  }
}
```

Now, about the instruction. The title should be about error completion, or observable completion, because the operator catchError will be used to avoid this observable completion. 

Then I like to start by explaning what the application is doing and how to use it. And then a small chapter on the goal of this challenge like your did. But what is misleading in your instruction is that you are taking of the right placement of catchError but you are not using it inside the challenge. You should start I think about taking about that the observable complete when you enter a wrong value. And the goal is to use the catchError operator to get a proper behavior. 

One more thing is that your code will create a memory leak. I think it should be better to add a takeUntilDestroy() because the challenge is not about memory leak.

DeveshChau:
Thanks, I will improve challenge with this guideline.

apps/rxjs-catch-error/src/app/app.component.ts
```diff
@@ -1,74 +1,45 @@
...
+  private destroyRef = inject(DestroyRef);
   constructor(private http: HttpClient) {}
-
   ngOnInit() {
-    const buttonClick$ = fromEvent(this.buttonRef.nativeElement, 'click');
-    buttonClick$
+    this.submit$$
       .pipe(
-        map(() => this.inputRef.nativeElement.value),
+        takeUntilDestroyed(this.destroyRef),
```
tomalaforge, line 41: nit: should be placed as the last operator so all obs are destroyed even the one inside high order operators. In this case, this is not such a big issue since http request completes on their own, but can lead to tricky bug otherwise

docs/src/content/docs/challenges/rxjs/38-rxjs-catch-error.md
```diff
@@ -1,25 +1,30 @@
...
-Handle the error without completion of the subscription.
+## Bug
+
+A bug has been identified in our application. Users are only able to successfully fetch data until an invalid request is sent. Once an error response is received, users are unable to send additional requests.
+
+## Learnings
+
+This application provides an opportunity to understand the correct placement of a `catchError` operator. If placed incorrectly, the overall subscription will be completed, preventing users from sending more requests. The goal is to preserve the overall subscription by handling error notifications from inner observables appropriately.
 
 ## Constraints
```
tomalaforge, line 25: That's way better. Just to make it more consistent with all other guideline, Can you create a structure like 
## Information
### How to use the application
### Bug
### Learnings

## Statement
The goal is to use the `catchError` operator to handle error management inside your Rxjs stream.

## Contraints

docs/src/content/docs/challenges/rxjs/38-rxjs-catch-error.md
```diff
@@ -1,25 +1,30 @@
 ---
 title: 🟢 catchError
-description: Challenge 38 is about learning error handling with catchError rxjs operator
+description: Challenge 38 is about learning obervable completion.
```
tomalaforge, line 3: You need to rebase your branch to the latest commit
The doc has changed
- you need to add some properties (look at another doc)
- delete l.9 
- delete all the footer

apps/rxjs-catch-error/src/app/app.component.css
```diff
@@ -0,0 +1,32 @@
...
+    border-radius: 4px;
+    cursor: pointer;
+  }
+  .response {
+    margin-left: 25%;
+    margin-top: 2%;
+    width: 50%;
+    text-align: center;
+    border: 1px solid #ccc;
+  }
```
tomalaforge, line 32: thanks, make the useful code more clean 👍

tomalaforge:
Thanks @DeveshChau. I added a few more comments but this si almost ready to go

tomalaforge:
@all-contributors add @DeveshChau for challenge

tomalaforge:
https://github.com/all-contributors add @DeveshChau  challenge

tomalaforge:
@all-contributors add @DeveshChau challenge

tomalaforge:
@DeveshChau I corrected the last small mistake but this is all good for me. Thanks a lot for your challenge. I'll post it this afternoon

tomalaforge (approved):

PR #196, opened 2023-10-14
Answer:9 (Wrap function pipe)

lalit-s-rajput:


apps/pipe-intermediate/src/shared/wrapFn.pipe.ts
```diff
@@ -0,0 +1,12 @@
...
+@Pipe({ name: 'wrapFn', standalone: true })
+export class WrapFunction implements PipeTransform {
+  transform(
+    callback: CallbackFunction,
+    arg1: string | number,
+    arg2: number | boolean
+  ) {
+    return callback(arg1 as string | number, arg2 as number | boolean);
+  }
+}
```
tomalaforge, line 12: Sorry but that's not very flexible. What if the function has 4 parameters ? What if the first one is an object or a boolean, or anything else ? 

this challenge is a big more complex. You should look at Typescript utility types to create more generic types

tomalaforge:
Thanks for submitting your work. However your solution will only work for this specific usecase. The challenge is too create a wrapper pipe to any possible functions

PR #197, opened 2023-10-14
Answer:10 (Pipe hard)

lalit-s-rajput:


tomalaforge:
Same as the one before, this challenge is more difficult because you need deep Typescript knowledge. The goal is to create a pipe that works for any use case. 
If you have any questions, I'll be happy to help you.

PR #200, opened 2023-10-15
Answer:22 (Router Input)

reed756:
Refactored ActivatedRouter with Input instead.


tomalaforge:
Awesome work

reed756:
Thanks for your review!

PR #201, opened 2023-10-16
Answer:31 (Module-to-Standalone)

reed756:
- Refactored app to use standalone components and bootstrapApplication.


tomalaforge:
Hello @reed756 , is the PR complete ? lots of files are missing

reed756:
Hey @tomalaforge, from the challenge description I wasn't too sure what to do, I refactored the app component to use standalone and also reconfigured the main.ts file to use bootstrapApplication. Were we also supposed to go into the MainShellModule and bring that into the app as standalone components, or something else? Thanks

reed756:
@tomalaforge I have just pushed a commit to update all components in the lib folder to standalone now.

tomalaforge:
That's perfect. 

you can also use the default keyword to avoid using the `then` function inside your import promise

PR #205, opened 2023-10-18
fix: checkbox challenge number

jdegand:
The links were broken so I investigated and challenge number was wrong.


tomalaforge (approved):

tomalaforge:
@all-collaborators add @jdegand for bug

PR #212, opened 2023-10-19
Answer 1: Projection task

mukhammadyor17:


apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,19 @@
...
   [class]="customClass">
-  <img
-    *ngIf="type === CardType.TEACHER"
-    src="assets/img/teacher.png"
-    width="200px" />
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px" />
+  <ng-content select="[image]"></ng-content>
```
tomalaforge, line 4: nit: you can also select the tag directly: `select=img`: this way you don't have to create a div. Or you can add the image selector directly on the img DOM element

mukhammadyor17: I fix that. thanks for comments sir.

PR #217, opened 2023-10-22
Answer: 1

DimitriTsikaridze:
Finished the challenge but I think  there are areas for improvement, mainly I don't like putting styles in global scope and
also challenge was about content projection, I used ng content but not ngTemplateOutlet didn't find any use case for it.

**Feedback on Challenge**

I have completed the challenge, but I believe there are areas for improvement:

1. **Global Styles:** I don't like putting styles in the global scope. couldn't think of anything without using ng deep.

2. **Different list names** I think I solved issue of list item names, (city having name and student/teacher firstname properties) in somewhat weird fashion.

3. **Content Projection:** While the challenge was about content projection, I used `ng-content` but didn't find a use case for `ngTemplateOutlet`.

any feedback would be much appreciated

### DIFF
--- a/apps/angular/projection/src/app/ui/card/card.component.html
+++ b/apps/angular/projection/src/app/ui/card/card.component.html
@@ -6,9 +6,8 @@
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item[itemNameKey]"
       [id]="item.id"
-      [type]="type"
       (onDelete)="onDelete.emit($event)">
     </app-list-item>
   </section>
--- a/apps/angular/projection/src/app/ui/card/card.component.html
+++ b/apps/angular/projection/src/app/ui/card/card.component.html
@@ -8,13 +8,14 @@
       *ngFor="let item of list"
       [name]="item.firstname"
       [id]="item.id"
-      [type]="type">
+      [type]="type"
+      (onDelete)="onDelete.emit($event)">
     </app-list-item>
   </section>
 
   <button
     class="border border-blue-500 bg-blue-300 p-2 rounded-sm"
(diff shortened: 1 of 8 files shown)


DimitriTsikaridze:
I didn't want to look at other solutions before I finished on my own, looks like I should've used ng-template inside parent component with list item component inside it and templateOutlet inside card component, great use case now that I think about  it.

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,20 @@
...
+  <ng-content select="img"></ng-content>
 
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item[itemNameKey]"
       [id]="item.id"
-      [type]="type">
+      (onDelete)="onDelete.emit($event)">
```
tomalaforge, line 11: as you mention inside your description, you should use ngTemplateOutlet here. I should have asked for more difference between cards because in this example, it's too simple to concatenate string to create a correct name.
But what happens if you need to add an icon. or anything else. Your shared component became less customizable.

apps/angular/projection/src/styles.scss
```diff
@@ -1,3 +1,15 @@
...
+  background-color: rgba(0, 250, 0, 0.1);
+}
+
+.bg-light-red {
+  background-color: rgba(250, 0, 0, 0.1);
+}
+
+.bg-light-blue {
+  background-color: rgb(30, 144, 255);
+}
```
tomalaforge, line 15: try with css variables for better customization

tomalaforge:
Nice work but your component is not easily customizable

PR #225, opened 2023-10-24
fix: add selectors for phone & video

jdegand:
Without selectors, angular can't tell the components apart - name collision warning in console.


tomalaforge:
This is weird, I don't have this warning in the console and never had an issue with that. If this was really an issue, selector won't be optional

Where do you have the warning?

tomalaforge (approved):

PR #226, opened 2023-10-24
Answer:39 Injection Token

jdegand:
This may not be right.   

There is a [warning](https://angular.io/errors/NG0912) in the console about name collision.  Both components are wrapped in ng-component.  

The one timer is slower than the other.  Both timers still go up by 1.   There is a delay of 1 and 2 seconds of the timer starting as well.  

I thought about eliminating the timer component but I decided to leave it.  I was tempted to refactor this as sort of an alternative solution.  The nesting prevents simple refactoring.


apps/angular/injection-token/src/app/video.component.ts
```diff
@@ -1,9 +1,12 @@
 import { Component } from '@angular/core';
 import { TimerContainerComponent } from './timer-container.component';
+import { TIMER_TOKEN } from './data';
 
 @Component({
+  selector: 'app-video',
   standalone: true,
   imports: [TimerContainerComponent],
+  providers: [{ provide: TIMER_TOKEN, useValue: 1000 }],
```
tomalaforge, line 9: TIMER_TOKEN should have a default value to 1000, so this override should be avoided

apps/angular/injection-token/src/app/timer.component.ts
```diff
@@ -1,13 +1,15 @@
...
+//import { DEFAULT_TIMER } from './data';
 
 @Component({
   selector: 'timer',
   standalone: true,
   template: ` Timer running {{ timer() }} `,
 })
 export class TimerComponent {
-  timer = toSignal(interval(DEFAULT_TIMER));
+  constructor(@Inject(TIMER_TOKEN) public time: number) {}
```
tomalaforge, line 13: perfer the `inject` function than the decorator. This way the property is properly typed

apps/angular/injection-token/src/app/phone.component.ts
```diff
@@ -1,9 +1,16 @@
 import { Component } from '@angular/core';
 import { TimerContainerComponent } from './timer-container.component';
+import { TIMER_TOKEN } from './data';
+
+// Phone and Video component don't have selectors - this causes collison issues
+// Both use ng-component
```
tomalaforge, line 6: I merged your PR. thanks for it

PR #229, opened 2023-10-26
feat(challenge38): solution

DeveshChau:


apps/rxjs/catch-error/src/app/app.component.ts
```diff
@@ -40,7 +40,11 @@ export class AppComponent {
       .pipe(
         map(() => this.input),
         concatMap((value) =>
-          this.http.get(`https://jsonplaceholder.typicode.com/${value}/1`)
+          this.http
+            .get(`https://jsonplaceholder.typicode.com/${value}/1`)
+            .pipe(
+              catchError((err) => of(console.log('Could not fetch api', err)))
```
tomalaforge, line 46: I don't understand why you are returning a console.log ? Never seen that before

DeveshChau: My aim is to handle error without sending error or complete notification to outer subscription.

PR #230, opened 2023-10-26
Answer:5

pchessah:


apps/angular/crud/src/app/components/todo-item/todo-item.component.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+      fixture.debugElement.nativeElement.querySelector('.update-btn');
+
+    expect(mockTodoItem.title).toBe('Original Title');
+
+    button.click();
+    fixture.detectChanges();
+
+    setTimeout(() => {
+      expect(mockTodoItem.title).toBe('New Title');
+    }, 2000);
```
tomalaforge, line 39: why do you need to use a setTimeout inside a test ? Your test will always take 2s to execute.

apps/angular/crud/src/app/components/todos/todos.component.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+      // Check that the todos are rendered
+      const todoElements =
+        fixture.nativeElement.querySelectorAll('.single-todo');
+      expect(todoElements.length).toBe(2);
+
+      // You can also assert the text content of individual todo items
+      expect(todoElements[0].textContent).toContain('Todo 1');
+      expect(todoElements[1].textContent).toContain('Todo 2');
+    }, 3000);
+  });
```
tomalaforge, line 46: same here, use FakeAsync, or done()

apps/angular/crud/src/app/models/todo-state.model.ts
```diff
@@ -0,0 +1,5 @@
+import { ITodo } from './todo.model';
+
+export interface ITodoState {
```
tomalaforge, line 3: don't add the letter I in front of your interface, bad practise. We already know that it's a type.

apps/angular/crud/src/app/components/todo-item/todo-item.component.ts
```diff
@@ -0,0 +1,94 @@
...
+          Update
+        </button>
+        <button
+          class="delete-btn"
+          [disabled]="!!todoItem.disabled"
+          (click)="deleteTodo()">
+          Delete
+        </button>
+      </div>
+    </div>
```
tomalaforge, line 23: loading should be local and not global when you update your item

apps/angular/crud/src/app/components/todos/todos.component.ts
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    this.todos$.pipe(take(1)).subscribe((todos) => {
+      const updatedTodos = todos.map((t) =>
+        t.id === updatedTodo.id ? updatedTodo : t
+      );
+      this.todoStore.loadTodos(updatedTodos);
+      const sub = this.todoService.update(updatedTodo);
+      this._subscriptions.push(sub);
+    });
```
tomalaforge, line 90: this code should be located inside your TodoStore. 

Service handle logic, component only display

apps/angular/crud/src/app/components/todos/todos.component.ts
```diff
@@ -0,0 +1,109 @@
...
+  loadingService = inject(LoadingService);
+  todoStore = inject(TodoStore);
+
+  todos$ = this.todoStore.todos$;
+  private _subscriptions: Subscription[] = [];
+
+  ngOnInit(): void {
+    this.todoService.get();
+    const todoSub$ = this.todoService.todos$.subscribe((todos) => {
+      this.todoStore.loadTodos(todos);
```
tomalaforge, line 69: load todo inside your store.

apps/angular/crud/src/app/components/todos/todos.component.ts
```diff
@@ -0,0 +1,109 @@
...
+        height: 100%;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+      }
+    `,
+  ],
+})
+export class TodosComponent implements OnInit, OnDestroy {
+  todoService = inject(TodoService);
```
tomalaforge, line 59: this should be a state of your store. Not a global state.

apps/angular/crud/src/app/error-handling/global-error-handler.ts
```diff
@@ -0,0 +1,14 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class GlobalErrorHandler implements ErrorHandler {
+  loadingService = inject(LoadingService);
+
+  handleError(error: any): void {
+    console.error('The following error was found:', error);
+    window.alert('There was an error: ' + error);
+    window.location.reload();
+    this.loadingService.stopLoading();
```
tomalaforge, line 12: this is a very dangerous way of handling errors. If a request never success, your application will reload infinitly. 

Why not creating an interceptor to log error, but don't put any logic here. In an entreprise application, errors might be handle differently.

tomalaforge:
Thanks a lot for your work. That's a good start, but your state is a big messy, and there is some state a big everywhere. Try to have your state located inside a service and call that service from your component to access your state. 

TodoItem should have a local state to handle loading for their own rows. 

If you have any quesitons, you can asked them here. You are on the right track by doing the challenge. Keep going. 🚀

apps/angular/crud/src/app/components/todos/todos.component.spec.ts
pchessah: @tomalaforge Could you show me how ? 
I had done the setTimeOut so that it waits for the todos on the screen. This is how I have used fakeAsync.

apps/angular/crud/src/app/error-handling/global-error-handler.ts
pchessah: @tomalaforge Instead of reloading the page, what about if I route to an error page showing the error. This page can have possible actions the user can take. I added this class for errors that are not http related.

tomalaforge: Yes this is better. You can create a 404 page inside your routing config and route to this page when there are some errors

PR #235, opened 2023-10-28
Answer;1

devanandTech:
subject:This is an optional longer description of the changes you made in this commit.


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -21,14 +26,18 @@ export class CardComponent {
...
   ) {}
 
   addNewItem() {
+    console.log(this.list);
     if (this.type === CardType.TEACHER) {
       this.teacherStore.addOne(randTeacher());
     } else if (this.type === CardType.STUDENT) {
       this.studentStore.addOne(randStudent());
+    } else if (this.type === CardType.CITY) {
+      this.cityStore.addOne(randomCity());
```
tomalaforge, line 40: This component should not have any logic inside. All the logic should be handle by the parent. Otherwise, everytime you need to add a new card, you will have to change this component. Not ideal. 

Look at ngTemplateOutlet 😉

PR #240, opened 2023-10-29
docs: injection token blogLink

jdegand:
I think your article 'stop being scared of injection tokens' should be added either as a hint or as a blogLink to the injection token docs.


tomalaforge (approved):

tomalaforge:
That's the great idea. But I think it should be added at a link inside the hint. The button at the button is more to answer the challenge. This article will be more about helping people on how to resolve it. 
What do you think?

tomalaforge (approved):
Perfect

tomalaforge:
Thanks a lot 🙏

PR #243, opened 2023-10-30
Answer:1

ygarciavillar:
# Content projection solution

I used a structural directive to provide the content projected and lead the ability to the type check in the template thanks to the ngTemplateContextGuard. My main struggle is trying to avoid one of the two issues.
1- Passing the same input values to the Card Component and the Directive to use the typing advantage or
2-Using OnPush Strategy on the Card Component with my current solution


tomalaforge:
I never thought of using a structural directive for this purpose. Indeed, you don't repeat the inputed value twice to have strong typing but it feels more complicated and harder to understand. 

And I cannot figure out why OnPush doesn't work on your PR.

ygarciavillar:
100% agree I only used the structural directive here like a sugar syntax..because I wanted to have type checking and I don't know if I can have this type checking without using the ngTemplateContextGuard. 

 <app-card class="bg-light-green" [listItems]="students$ | async" (addItem)="addStudent()">
      
      **<ng-template [cardContent]="students$ | async" let-student>
          <app-list-item (delete)="deleteStudent(student.id)">
            {{ student.firstname }}
          </app-list-item>
      </ng-template>**
   </app-card>

Thanks for these Angular challenges, I'm enjoying it and trying to improve my Angular knowledge.

tomalaforge:
You can have a look at this [article](https://medium.com/@thomas.laforge/ngtemplateoutlet-type-checking-5d2dcb07a2c6). But you will just have two inputs instead of combining the inputed list.

jdegand:
> I never thought of using a structural directive for this purpose. Indeed, you don't repeat the inputed value twice to have strong typing but it feels more complicated and harder to understand.
> 
> And I cannot figure out why OnPush doesn't work on your PR.

You can't use OnPush with directives directly.  You would need to use a wrapper component with the directive component inside.

PR #246, opened 2023-11-02
Answer:14

reed756:
Added startWith operator

### DIFF
--- a/apps/rxjs-race-condition/src/app/app.component.ts
+++ b/apps/rxjs-race-condition/src/app/app.component.ts
@@ -19,7 +19,7 @@ export class AppComponent {
   title = 'rxjs-race-condition';
   dialog = inject(MatDialog);
   topicService = inject(TopicService);
-  topics$ = this.topicService.getTopics();
+  topics$ = this.topicService.topics$;
 
   openTopicModal(topics: { topics: TopicType[] }) {
     this.dialog.open(TopicModalComponent, { data: topics });
--- a/apps/rxjs-race-condition/src/app/app.component.ts
+++ b/apps/rxjs-race-condition/src/app/app.component.ts
@@ -1,38 +1,27 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  inject,
-  OnInit,
-} from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { TopicModalComponent } from './topic-dialog.component';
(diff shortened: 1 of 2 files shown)


apps/rxjs-race-condition/src/app/app.component.ts
```diff
@@ -19,12 +19,12 @@ export class AppComponent implements OnInit {
...
   dialog = inject(MatDialog);
   topicService = inject(TopicService);
-  topics: TopicType[] = [];
+  topics: TopicType[] | string[] = [];
 
   ngOnInit(): void {
     this.topicService
       .fakeGetHttpTopic()
-      .pipe(take(1))
+      .pipe(take(1), startWith(['Politic', 'Culture', 'Nature']))
```
tomalaforge, line 27: That's not the solution. Indeed you will get data inside your topics array, but not the right one. 

The goal is to see that imperative coding can create race condition. Refactor the code to declarative coding.

reed756: Thanks, I'll have another go

reed756: @tomalaforge Hey I've done a new commit which has a declarative solution now.

apps/rxjs-race-condition/src/app/topic.service.ts
```diff
@@ -1,10 +1,29 @@
...
+      const data = {
+        topics: topics,
+      };
+      this.topics$.next(data);
+    });
+  }
+
+  getTopics() {
+    return this.topics$;
+  }
```
tomalaforge, line 28: nit: prefer to return the observable and not the subject. 

The subject should not be accessible from the outside. 

Donc create a method, 

just do 
```ts
topics$ = this.topics$$.asObservable()
```

And use $$ or nothing for subject

apps/rxjs-race-condition/src/app/topic.service.ts
```diff
@@ -1,10 +1,29 @@
...
+    this.init();
+  }
+
   fakeGetHttpTopic = () =>
     timer(1000).pipe(map((): TopicType[] => ['Politic', 'Culture', 'Nature']));
+
+  init() {
+    this.fakeGetHttpTopic().subscribe((topics) => {
+      const data = {
+        topics: topics,
```
tomalaforge, line 20: nit, when the property and the value have the same name, you don't have to repeat it.

tomalaforge:
This is way better. Nice job

reed756:
Thank you for the review and the pointers to make it better, much appreciated!

PR #256, opened 2023-11-10
Answer:1 Projection

hettiger:
I successfully removed `::ng-deep` while maintaining the styles. However, I'm not sure if I did it in the way that I was supposed to do it. (I used host on the CardComponent Component decorator and removed the div so that the styles apply)

### DIFF
--- a/apps/angular/projection/src/app/component/student-card/student-card.component.ts
+++ b/apps/angular/projection/src/app/component/student-card/student-card.component.ts
@@ -9,11 +9,12 @@ import { CardComponent } from '../../ui/card/card.component';
 import { NgForOf, NgIf } from '@angular/common';
 import { CardItemDirective } from '../../ui/card/card-item.directive';
 import { ListItemComponent } from '../../ui/list-item/list-item.component';
+import { toSignal } from '@angular/core/rxjs-interop';
 
 @Component({
   selector: 'app-student-card',
   template: `<app-card
-    [list]="students"
+    [list]="students()"
     (add)="addNewItem()"
     class="bg-light-green">
     <img appCardImage src="assets/img/student.webp" width="200px" />
@@ -34,14 +35,12 @@ import { ListItemComponent } from '../../ui/list-item/list-item.component';
   imports: [CardComponent, NgIf, CardItemDirective, ListItemComponent, NgForOf],
 })
 export class StudentCardComponent implements OnInit {
-  students: Student[] = [];
+  students = toSignal(this.store.students$, { initialValue: [] as Student[] });
 
   constructor(private http: FakeHttpService, private store: StudentStore) {}
 
   ngOnInit(): void {
     this.http.fetchStudents$.subscribe((s) => this.store.addAll(s));
-
-    this.store.students$.subscribe((s) => (this.students = s));
   }
 
   addNewItem() {
--- a/apps/angular/projection/src/app/component/student-card/student-card.component.ts
+++ b/apps/angular/projection/src/app/component/student-card/student-card.component.ts
@@ -13,10 +13,7 @@ import { toSignal } from '@angular/core/rxjs-interop';
 
 @Component({
   selector: 'app-student-card',
-  template: `<app-card
-    [list]="students()"
-    (add)="addNewItem()"
-    class="bg-light-green">
+  template: `<app-card [list]="students()" (add)="addNewItem()">
     <img appCardImage src="assets/img/student.webp" width="200px" />
     <ng-template appCardItem let-item>
       <app-list-item (delete)="delete(item.id)">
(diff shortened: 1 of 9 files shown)


apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,13 @@
...
-  [class]="customClass">
-  <img
-    *ngIf="type === CardType.TEACHER"
-    src="assets/img/teacher.png"
-    width="200px" />
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px" />
+<ng-container *ngTemplateOutlet="cardImage.templateRef"></ng-container>
```
tomalaforge, line 1: When there is no logic to give back to the parent or the template you need to project, use ng-content. This is way easier

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,17 @@
...
+import { Component, EventEmitter, Output } from '@angular/core';
 
 @Component({
   selector: 'app-list-item',
   template: `
     <div class="border border-grey-300 py-1 px-2 flex justify-between">
-      {{ name }}
-      <button (click)="delete(id)">
+      <ng-content></ng-content>
+      <button (click)="delete.emit()">
```
tomalaforge, line 8: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,29 @@
...
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @ContentChild(CardImageDirective) cardImage!: CardImageDirective;
+  @ContentChild(CardItemDirective) cardItem!: CardItemDirective;
```
tomalaforge, line 28: Using a directive is the best way to do it. nice job 👍

If you want, you can use the property read of ContentChild to create a TemplateRef, instead of using the contructor of the directive, but both way work perfectly

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -32,4 +52,12 @@ export class TeacherCardComponent implements OnInit {
 
     this.store.teachers$.subscribe((t) => (this.teachers = t));
```
tomalaforge, line 53: You should refactor this, either using rxjs or now signals

tomalaforge:
About the style, in this example, ng-deep is easy to delete. I wanted people to use css variable but it's useless in this scenario. That is why I created another challenge about styling.

hettiger:
I've refactored everything as suggested.
I also did bring back the root div in the card component to show that it works in this case too.

Thank you so much for taking the time @tomalaforge really appreciate it!

apps/angular/projection/src/styles.scss
```diff
@@ -1,3 +1,7 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+:root {
+  --app-card-background-color: transparent;
```
tomalaforge, line 6: nice, design token are the best 🔥

apps/angular/projection/src/app/component/student-card/student-card.component.ts
```diff
@@ -1,35 +1,44 @@
...
-    `,
-  ],
-  imports: [CardComponent],
+  styles: [':host { --app-card-background-color: rgba(0, 250, 0, 0.1); }'],
+  imports: [CardComponent, NgIf, CardItemDirective, ListItemComponent, NgForOf],
 })
 export class StudentCardComponent implements OnInit {
-  students: Student[] = [];
-  cardType = CardType.STUDENT;
+  students = toSignal(this.store.students$, { initialValue: [] as Student[] });
```
tomalaforge, line 29: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,29 @@
...
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @ContentChild(CardItemDirective, { read: TemplateRef })
+  itemTemplate!: TemplateRef<{ $implicit: T }>;
```
tomalaforge, line 28: 👍

tomalaforge:
Perfect 🔥

hettiger:
> Perfect 🔥

Thank you for the feedback and your time!

PR #258, opened 2023-11-11
Answer:1 Projection

Brack0:

## Notes

- Missing some types on `let-item` but there is challenge 4 for this one 😅
- TemplateContext with `$implicit` is a nice trick


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,55 @@
...
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
+    this.addItem.emit();
+  }
+
+  trackById(_index: number, item: ListItem) {
+    return item.id;
```
tomalaforge, line 53: in a few days, you will be able to use the new control flow with the new track function 👍 but nice to have implement it, not everyone think of it.

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,35 +1,60 @@
...
       }
     `,
   ],
   standalone: true,
-  imports: [CardComponent],
+  imports: [
+    CardComponent,
+    ListItemComponent,
+    ListItemTemplateDirective,
+    AsyncPipe,
```
tomalaforge, line 40: you can use the new signal API to get read of this. Signal will be the new way of binding properties to the template, it's better to get use to it now. And many benefits will come in a very near future

tomalaforge:
very nice work. you thought of everything. Maybe refacto to signals to make it perfect. 

And as your said, you could have type check the directive. But there is a challenge for that. 😉

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
Brack0: Yep, still in developer preview (as of v16) but could make use of it. Hoping that interop with RxJS is working as intented.

Brack0:
Cya for the next challenge !

PR #259, opened 2023-11-11
Answer:1 Projection

Rezorl:


apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,30 @@
...
       <button (click)="delete(id)">
         <img class="h-5" src="assets/svg/trash.svg" />
       </button>
     </div>
   `,
   standalone: true,
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class ListItemComponent {
   @Input() id!: number;
```
tomalaforge, line 23: id can be deleted, you only need to know when the user click on the button, and id can be pass inside the parent element

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,40 @@
...
+
+  @Output() addItem = new EventEmitter();
 
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore
-  ) {}
+  @ContentChild(TemplateRef, { static: true }) templateRef: TemplateRef<{
+    item: T;
+  }> | null = null;
```
tomalaforge, line 31: you should name your template inside the parent, using a directive. If more template get added, you may run into troubles

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,21 +1,14 @@
...
 
   <section>
     <app-list-item
       *ngFor="let item of list"
-      [name]="item.firstname"
       [id]="item.id"
-      [type]="type">
+      (deleteItem)="delete($event)">
+      <ng-template [ngTemplateOutlet]="templateRef" [ngTemplateOutletContext]="{item: item}"/>
     </app-list-item>
```
tomalaforge, line 12: to make it even more customizable, the list item can be call inside the parent. Because what will happen if we want to create a card with list item without a delete button ?

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,35 +1,51 @@
...
+  </app-card>`,
   standalone: true,
-  imports: [CardComponent],
+  imports: [CardComponent, NgIf],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class TeacherCardComponent implements OnInit {
-  teachers: Teacher[] = [];
-  cardType = CardType.TEACHER;
+  teachers = toSignal<Teacher[], Teacher[]>(this.store.data$, {
```
tomalaforge, line 34: nice signals 😍

tomalaforge:
very good job and very nice refacto using the latest features. However be careful with your contentChild, you should always target an element inside your parent using a directive. It's safer

Rezorl:
> very good job and very nice refacto using the latest features. However be careful with your contentChild, you should always target an element inside your parent using a directive. It's safer

ok, thanks for review

PR #264, opened 2023-11-11
Answer: 39

ygarciavillar:
I like this Angular challenges series. It is a pity that the Angular QA/coding of the past Friday didn't read the README properly, but it is such a great idea this Angular challenge with the Nx repo included. Congrats!.

### DIFF
--- a/apps/angular/injection-token/src/app/app.config.ts
+++ b/apps/angular/injection-token/src/app/app.config.ts
@@ -1,5 +1,11 @@
-import { ApplicationConfig } from '@angular/core';
+import { ApplicationConfig, InjectionToken } from '@angular/core';
 import { provideRouter } from '@angular/router';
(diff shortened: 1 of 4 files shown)


tomalaforge:
thanks 😍 and I agree with you, I was so frustrated with the stream of the angular team. I spend a tremendous amount of time creating a documentation 😭

tomalaforge:
Perfect answer, great job 🎉

ygarciavillar:
Yes... I was so frustrating for me too..I couldn't believe it because the documentation is so clear..they only need 5 min or less to read it. But anyway excellent job.

tomalaforge:
Yes thanks for your feedback. But you can see that even the best devs don't know how to read a documentation 😅

ygarciavillar:
Yes, you are right!!🤣🤣

PR #269, opened 2023-11-12
Answer:5

DragosAGrigore:
Done the first 2 steps, will come to the next steps in the future.

### DIFF
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,51 +1,32 @@
 import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
-import { randText } from '@ngneat/falso';
+import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
+import { Todo } from './todo.model';
+import { TodoService } from './todo.service';
+import { Observable } from 'rxjs';
 
 @Component({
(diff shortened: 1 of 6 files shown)


apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,44 @@
...
+      <button (click)="delete(todo.id)">Delete</button>
     </div>
   `,
   styles: [],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class AppComponent implements OnInit {
-  todos!: any[];
+  public todos$: Observable<Todo[]> = this.todoService.todos$;
+  public loader$: Observable<boolean> = this.loaderService.loader$;
```
tomalaforge, line 26: you can use signal now 😜

apps/angular/crud/src/app/error.interceptor.ts
```diff
@@ -0,0 +1,29 @@
...
+    return next.handle(req).pipe(
+      finalize(() => this.loaderService.hideLoader()),
+      catchError((error) => {
+        console.log(error);
+        this.loaderService.hideLoader();
+        return of();
+      })
+    );
+  }
+}
```
tomalaforge, line 29: my personal opinion: this seems great for small application but it will always trigger a loader for http request, and sometimes you don't want to. 

For instance, on this example, we want to trigger a loader per TodoItem when we update it,

PR #273, opened 2023-11-12
Answer:2 Effect vs Selector

Brack0:

## Notes

- I discovered and used createActionGroup.
- Not that hard IMHO.
- Any strong opinion on this ? _Only one action should be dispatched from a component._ Is this just a rule for the challenge or should it be applied in projects ? It sounds like a good idea but not sure how to justify this.
- ActivityCardComponent is not really reusable :/


tomalaforge:
Very great work. It's not hard, but many people don't understand what to put where. When you understand the mental model of ngrx/store, everything feels natural. 

About single action name per component: this is a very important rule. If you use the redux devtool and you want to debug your application, you will see how events flow inside your application very easily. If 5 different components dispatch the same action, good luck to find where the bug come from. 
When the application is small, that's ok, but when you will have 100 of components, this simple rule can save you.

PR #276, opened 2023-11-12
Answer:8 - pipe easy

DragosAGrigore:


apps/angular/pipe-easy/src/app/heavy-computation.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'heavComputation',
```
geromegrignon, line 4: There is a typo in the pipe name: it should be `heavyComputation`

geromegrignon (changes requested):

apps/angular/pipe-easy/src/app/heavy-computation.pipe.ts
DragosAGrigore: Yes, thank you!

PR #277, opened 2023-11-12
Answer:3 Directive Enhancement

Brack0:

## Notes

My first attempt to solve this challenge was to extends NgFor (`export class NgForEmptyDirective<T, U> extends NgForOf`) and use directive composition to plug `NgIf`. And then I remember that extends is often a bad idea, I had to deal with many stuff from NgFor (inheritance hell).
On my second try, I swaped directives (NgFor in hostDirectives , and extends NgIf). Then I realized that ngIf is just a toggle on a viewContainer, so I can do it myself instead of extending the real directive.

I might be missing some cases here (such as dynamic swapping of the EmptyTemplate).
Performance should be ok, ngOnChanges is linked to Input properties and we can assume we won't have many reference changes in ngFor.

EDIT : Third challenge in a row 🎉

### DIFF
--- a/apps/angular/ngfor-enhancement/src/app/directives/ng-for-empty.directive.ts
+++ b/apps/angular/ngfor-enhancement/src/app/directives/ng-for-empty.directive.ts
@@ -1,0 +1,33 @@
+import { NgFor } from '@angular/common';
+import {
+  Directive,
+  Input,
+  OnChanges,
+  TemplateRef,
+  ViewContainerRef,
+  inject,
+} from '@angular/core';
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: '[ngFor][ngForOf]',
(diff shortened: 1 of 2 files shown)


tomalaforge:
Thanks for the description and very nice solution. 👍

apps/angular/ngfor-enhancement/src/app/directives/ng-for-empty.directive.ts
othAs-dev, line 1: this code doesn't works correctly when the list is more one element i tested it and when i have more one element in my persons list the template will render only the first item of the list and not the second :)

Brack0: True, `this.viewContainer.remove();` was not necessary and caused issues with the list (one item from the list was removed)

PR #278, opened 2023-11-12
feat(docs): add internationalization and add spanish lang

nelsongutidev:
This PR adds internationalization to the docs site and adds Spanish as the first option for an additional language. 

Internationalization is added following the guidelines on the Starlight official docs. I added a root locale (english) so all links continue to work as they do now. This just sets up everything so additional languages can be added. 

As a part of this work, I translated the main page to Spanish as well as all the guide sections and 2 angular challenges. I can work on the rest of the translations for Spanish if you decide to go forward with this. Please feel free to decline this PR if you decide not to go in this direction, it is completely fine and understandable too.

Starlight is awesome in the sense that untranslated pages will fall back to the English ones so this should be a non-breaking change.

image
image

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -1,6 +1,11 @@
 import starlight from '@astrojs/starlight';
 import { defineConfig } from 'astro/config';
 
+export const locales = {
+  root: { label: 'English', lang: 'en' },
+  es: { label: 'Español', lang: 'es' },
+};
+
 // https://astro.build/config
 export default defineConfig({
   integrations: [
@@ -21,10 +26,16 @@ export default defineConfig({
         {
           label: 'Guides',
           autogenerate: { directory: 'guides' },
+          translations: {
+            es: 'Guías',
+          },
         },
         {
           label: 'Challenges',
           autogenerate: { directory: 'challenges' },
+          translations: {
+            es: 'Desafíos',
+          },
         },
       ],
       head: [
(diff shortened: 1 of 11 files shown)


tomalaforge:
This is amazing work. 😍 I'll review when I have more time. I'm currently traveling.

tomalaforge:
This is very good work. Thanks a lot. 
The only thing is that I will be unable to maintain it, and when I change something inside a page, it's hard to do the change in all locales. 
But since all translated file are separate files, it's still ok. 

thanks a lot and of course, I will merged it. 👍

tomalaforge:
@all-contributors add @nelsongutidev for doc

tomalaforge (approved):

nelsongutidev:
> This is very good work. Thanks a lot. The only thing is that I will be unable to maintain it, and when I change something inside a page, it's hard to do the change in all locales. But since all translated file are separate files, it's still ok.
> 
> thanks a lot and of course, I will merged it. 👍

Thanks man! I will do my best to keep it updated with the Spanish version. I have a WIP of the rest of the challenges, a PR should be up soon

tomalaforge:
thanks but be careful because I need to rewrite most of the challenge
documentation. but right now and until end of the year. I don't have a lot
of time.

Challenge without the note at the top should not change. You can start with
these one.

Thanks a lot for doing this. 👍

Le mer. 15 nov. 2023, 22:46, Nelson Gutierrez ***@***.***> a
écrit :

PR #280, opened 2023-11-13
Answer:4 Typed ContextOutlet

Brack0:

## Notes

- ngTemplateContextGuard is not smart enough to infer generic without an input. Not sure why this is happening, but this make more sense when it comes to shorthand syntax.
- Thanks for the reminder on how to query a template variable with `@ContentChild("#myRef")`. I could have used it for challenge 1 instead of a dummy directive.
- Now I want to use this in real life.

### DIFF
--- a/apps/angular/context-outlet-type/src/app/app.component.ts
+++ b/apps/angular/context-outlet-type/src/app/app.component.ts
@@ -15,9 +15,9 @@ import { PersonComponent, PersonTemplateDirective } from './person.component';
   selector: 'app-root',
   template: `
     <person [person]="person">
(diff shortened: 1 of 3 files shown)


tomalaforge:
That's perfect, nothing to add. 

- Yes you have to add an input, otherwise how can Angular know your type ? the type has to come from somewhere 😉
- It's not a reminder, it's bad practise. Prefer always using a dummy directive than magic string. It's safer. With magic string, you can have a typo and don't understand why you template if not seen by the child
- go for it, use it and spread the word. NgTemplateOutlet type checking is next level. You will impress your coworkers with that and make your application safer. 👍

PR #281, opened 2023-11-13
Answer 22: Angular router input

Clarchik:


package.json
```diff
@@ -92,7 +92,7 @@
     "postcss-import": "~14.1.0",
     "postcss-preset-env": "~7.5.0",
     "postcss-url": "~10.1.3",
-    "prettier": "^2.6.2",
+    "prettier": "^3.1.0",
```
geromegrignon, line 95: While important to update the dependencies of a project, a commit should focus on its goal. Such a change should land in a dedicated commit.

tomalaforge: you can create a PR if you want to upgrade prettier, but a global update will happen soon.

PR #282, opened 2023-11-13
Answer: 3

ygarciavillar:
Another excellent challenge.!


tomalaforge:
nice work, but this challenge will be soon obsolete for people upgrading to v17 🔥

PR #287, opened 2023-11-14
Answer:31 - module to standalone

DragosAGrigore:


tomalaforge:
That's perfect. Did you use the angular schematic ? 
And you can go one step further by exporting your routes and component as default to avoid the `then(c => c.xxx)` but this come to personal preferencies

DragosAGrigore:
No, I didn't use the angular schematic.
I forced myself to analyze and explore what needs to be changed in order to have a full standalone application.
Thank you for the response, I will surely apply that in my work.

tomalaforge:
you for sure should, standalone is the way to go now

PR #288, opened 2023-11-14
Answer:5

AlexLukian:


apps/angular/crud/src/app/services/todo.service.ts
```diff
@@ -0,0 +1,52 @@
...
+import { randText } from '@ngneat/falso';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class TodoService {
+  private readonly http = inject(HttpClient);
+
+  private todosList = new BehaviorSubject<Todo[]>([]);
+  public todoList$: Observable<Todo[]> = this.todosList.asObservable();
```
tomalaforge, line 14: You can use signals now 😉

And for better code separation, I think it's better to have your http request inside a separate service as your state. 

And setting the state inside the tap operator feel weird. I don't know if that's very maintainable on entreprise application, but that's my personal preference.

apps/angular/crud/src/app/interceptors/error-interceptor.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (error.error && error.error.message) {
+          errorMessage = error.error.message;
+        }
+
+        this.snackBar.open(errorMessage, 'Close', {
+          duration: 5000,
+          verticalPosition: 'top',
+        });
+
+        return EMPTY;
```
tomalaforge, line 34: 👍

apps/angular/crud/src/app/components/app.component.ts
```diff
@@ -0,0 +1,50 @@
...
+  private readonly destroyRef = inject(DestroyRef);
+
+  todosList$!: Observable<Todo[]>;
+
+  ngOnInit(): void {
+    this.todoService
+      .initTodos()
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe();
+    this.todosList$ = this.todoService.todoList$;
```
tomalaforge, line 34: Don't do that 😱
You want to do reactive programming and you are kind of mixing declarative and reactive. 

Inside your service, you are setting a subject, that you inject new value into it. The view will react to those value. Don't reset your observable. 

Just do `todosList$ = this.todoService.todoList$` l.27

apps/angular/crud/src/app/components/app.component.ts
```diff
@@ -0,0 +1,50 @@
...
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe();
+  }
+
+  onTodoDelete(todo: Todo) {
+    this.todoService
+      .deleteTodo(todo)
+      .pipe(takeUntilDestroyed(this.destroyRef))
+      .subscribe();
+  }
```
tomalaforge, line 49: it's nice to use new operator like `takeUntilDestroyed` but in your case, you are interesting in only the first emitted value so use `take(1)` because in your case, your observable will never get destroy (not really true because you are calling a http request with complete on its own but you need to be careful) and everytime you click on a button, a new observable will occur.

apps/angular/crud/src/app/services/todo.service.ts
AlexLukian: Would it be ok if i set the state not of the current service, but of different one inside the tap operator?

Or, if the problem is with using tap for state management, what is the better way to do it?

apps/angular/crud/src/app/components/app.component.ts
AlexLukian: Thank you for the insightful comment. I didn't consider that the destruction of the component may never occur. 
Also, as i see, you didn't comment on other people not unsubscribing from http requests. Isn't it causing memory leaks or in current angular versions there is no need for unsubscriptions from them?

tomalaforge: That's a very good remark. 
Indeed I don't comment other person PR because it's better practise to unsuscribe but it will not leak because http will complete when the request respond. 
The only thing that can happen is if you quit a component and the request take time, you can have weird side effect because the observable is not unsuscribe.

apps/angular/crud/src/app/services/todo.service.ts
tomalaforge: the best way is to have a state service that is calling your http service. But the http service is only dealing with the request. I think I prefer to have the subscribe inside the service instead of outside. That's many the thing that bother me.

AlexLukian: Thank you very much for your answers

apps/angular/crud/src/app/components/app.component.ts
AlexLukian: Thank you very much for your answers

apps/angular/crud/src/app/services/todo.service.ts
AlexLukian: > the best way is to have a state service that is calling your http service. But the http service is only dealing with the request. I think I prefer to have the subscribe inside the service instead of outside. That's many the thing that bother me.

If we subscribe within the service, then how can we unsubscribe? I looked at other answers and didn't find an example.

tomalaforge: The service is linked to a component life

AlexLukian: The service is created as a singleton and lives with the application all the time. Or I'm wrong ? Thanks for answers.

tomalaforge: If you don't provide it at the component level, you are right. 
You should read my last article
https://medium.com/p/cc210f7040bd

PR #289, opened 2023-11-14
Answer:3

Clarchik:


apps/angular/ngfor-enhancement/src/app/ng-for-enhancened.directive.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.ref?.destroy();
+    const isEmpty =
+      !(this as any)._ngForOf || (this as any)._ngForOf.length === 0;
+    if (isEmpty) {
+      this.ref = this.vcr.createEmbeddedView(this.ngForElse);
+    } else {
+      this.ref?.destroy();
+    }
+  }
+}
```
tomalaforge, line 47: you can indeed extend your directive, but v15 released the hostDirective property which let you enhance a directive without having to extend the contructor or making things more complex.

tomalaforge:
Your solution works but try with hostDirectives to try new features

tomalaforge:
that's way better, isn't it ?

PR #297, opened 2023-11-15
Answer:31

BruceArmstrong007:

  Converted Angular Project to Standalone (NgModule -> Standalone)


libs/module-to-standalone/admin/feature/src/lib/admin-feature.routes.ts
```diff
@@ -0,0 +1,16 @@
+import { Route } from '@angular/router';
+
+export const adminRoutes: Route[] = [
+  {
+    path: '',
+    loadChildren: () =>
+      import('./dashboard/dashboard.routes').then((m) => m.dashboardRoutes),
```
tomalaforge, line 7: you can use the default keyword to avoid `then`inside your promise

tomalaforge:
Very good 👍 Did you try to use the angular schematic for this ?

libs/module-to-standalone/admin/feature/src/lib/admin-feature.routes.ts
BruceArmstrong007: ya that will be much cleaner as well, thanks i will implement it from now

BruceArmstrong007:
> Very good 👍 Did you try to use the angular schematic for this ?

@tomalaforge  yes, i do use nx angular schematic for generating components, pipes and directives and stuff

PR #300, opened 2023-11-15
Answer: 6

Clarchik:


apps/angular/permissions/src/app/user.store.ts
```diff
@@ -9,6 +9,25 @@ export class UserStore {
...
+        }
+
+        if (user.isAdmin) {
+          return true;
+        }
+
+        const roles: Role[] = Array.isArray(role) ? role : [role];
+        return roles.length === 0 || user.roles.some((r) => roles.includes(r));
+      }),
+    );
```
tomalaforge, line 29: you can use signal now 😜

tomalaforge:
That's perfect, but to go a bit further and play with the signal API, I advice you to change everything to signal. There is no async state here, so you can fully embrace the signal reactivity system

apps/angular/permissions/src/app/user.store.ts
Clarchik: I tried to, but found it difficult to refactor directive using signal. May you show solution with singnals?
I don't know to listen/run login inside of directive while signal changes instead of current component store effect

PR #301, opened 2023-11-15
Answer:34 - Performance Default On Push

Eduardoooxd:
The only thing that I could improve is the fact that the title of the person list is being re rendered while focusing on the input, I could separate on a different component but I think it would be overkill for this case. 
Either way, there should be a way to prevent the re render of the title.

- [x] Start your PR message with Answer 34


tomalaforge:
Yes that's perfect and soon localchangedetection will be possible. We need to wait a bit longer.

PR #303, opened 2023-11-16
Answer:6

BruceArmstrong007:

Added and implemented canMatch guard & hasRole directive


apps/angular/permissions/src/app/common/guards/is-admin.guard.ts
```diff
@@ -0,0 +1,18 @@
...
+import { CanMatchFn } from '@angular/router';
+import { UserStore } from '../../user.store';
+import { map } from 'rxjs';
+import { TypedRoute } from '../../routes';
+
+export const HasPermission: CanMatchFn = (route: TypedRoute, segments) => {
+  const user$ = inject(UserStore).user$;
+  return user$.pipe(
+    map((user) => {
+      if (route?.data?.isAdmin) return user?.isAdmin ? true : false;
```
tomalaforge, line 11: don't need the true and false.

apps/angular/permissions/src/app/common/guards/is-admin.guard.ts
```diff
@@ -0,0 +1,18 @@
...
+      else {
```
tomalaforge, line 12: don't need the else, you already have a early return

apps/angular/permissions/src/app/routes.ts
```diff
@@ -1,14 +1,78 @@
-export const APP_ROUTES = [
+import { Route } from '@angular/router';
+import { Role } from './user.model';
+import { HasPermission } from './common/guards/is-admin.guard';
+export interface TypedRoute extends Route {
+  data?: {
+    isAdmin?: boolean;
+    roles?: Role[];
+  };
+}
```
tomalaforge, line 9: 👍

apps/angular/permissions/src/app/routes.ts
```diff
@@ -1,14 +1,78 @@
...
+export const APP_ROUTES: TypedRoute[] = [
   {
     path: '',
     loadComponent: () =>
       import('./login.component').then((m) => m.LoginComponent),
   },
   {
     path: 'enter',
+    canMatch: [HasPermission],
```
tomalaforge, line 18: HasPermission is a function which can take parameters. Don't use the data property

apps/angular/permissions/src/app/common/guards/is-admin.guard.ts
BruceArmstrong007: yes, it doesn't need to be , i will fix it

apps/angular/permissions/src/app/routes.ts
BruceArmstrong007: @tomalaforge  u mean i can call that function directly as well ?

tomalaforge: That's the goal of the new API. Guard are function, not class,

BruceArmstrong007: i will update that.

PR #306, opened 2023-11-16
Answer:32

BruceArmstrong007:

Removed function and added menu observable in async pipe in as input


apps/angular/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -40,23 +41,15 @@ export class NavigationComponent {
...
+  template: ` <app-nav [menus]="(menus$ | async) ?? []" /> `,
   host: {},
 })
 export class MainNavigationComponent {
   private fakeBackend = inject(FakeServiceService);
 
-  readonly info$ = this.fakeBackend.getInfoFromBackend();
+  readonly menus$ = this.fakeBackend
+    .getInfoFromBackend()
+    .pipe(map((info) => this.getMenu(info ?? '')));
```
tomalaforge, line 52: you can now use the signal API, but the solution is correct

BruceArmstrong007: to be fair, i thought about converting this http response observable toSignal,  but used rxjs since i had to transform the value by sending to getMenu function to get the result value for the template

PR #308, opened 2023-11-16
Answer:39

BruceArmstrong007:

Added injection token for timer and provided its usevalue in nested components


apps/angular/injection-token/src/app/video.component.ts
```diff
@@ -1,14 +1,22 @@
...
 @Component({
   selector: 'app-video',
   standalone: true,
   imports: [TimerContainerComponent],
+  providers: [
+    {
+      provide: DEFAULT_TIMER_TOKEN,
+      useValue: '1000'
+    }
+  ],
```
tomalaforge, line 14: there is a better solution. You can provide a default value to your injection token.

BruceArmstrong007: @tomalaforge  mybad i should have removed it , since the injection_tokens default value is already 1000

tomalaforge: no you don't have any default value

BruceArmstrong007: @tomalaforge  the Timer token from the data.ts  is injected in timerContainer component ,  and the video component will have no need for providing '1000' value, since its already injected in timerContainer component

tomalaforge: I don't understand what your are trying to say. The goal is to have a default value at 1000 and override it when needed. If you delete this provider, you will have an error

BruceArmstrong007: @tomalaforge  i have realized the syntax for setting default value in data.ts was wrong,
but ya after fixing that, i have to remove the provider from video compoent since it have same timer as the default value, just need to inject the injection_token in that component

apps/angular/injection-token/src/app/data.ts
```diff
@@ -1,6 +1,8 @@
...
+import { InjectionToken } from '@angular/core';
 
 export const DEFAULT_TIMER = '1000';
 
-export const DEFAULT_TIMER_TOKEN = new InjectionToken(DEFAULT_TIMER);
-
+export const DEFAULT_TIMER_TOKEN = new InjectionToken('TOKEN', {
+  providedIn: 'root',
+  factory: () => DEFAULT_TIMER,
+});
```
tomalaforge, line 8: perfect, that's how we set a default value. Great job 👍

PR #311, opened 2023-11-16
Answer: 9

Clarchik:


apps/angular/pipe-intermediate/src/app/wrap-fn.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'wrapFn',
+  standalone: true,
+})
+export class WrapFnPipe implements PipeTransform {
+  transform<T, F extends (...args: never[]) => T>(
```
tomalaforge, line 8: For my information: Why did you use never here ?

PR #314, opened 2023-11-16
Answer: 32

Clarchik:


apps/angular/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -35,30 +36,26 @@ interface MenuItem {
...
+  template: ` <app-nav [menus]="(menus$ | async)!"></app-nav> `,
   host: {},
 })
 export class MainNavigationComponent {
   private fakeBackend = inject(FakeServiceService);
 
-  readonly info$ = this.fakeBackend.getInfoFromBackend();
+  private readonly info$ = this.fakeBackend.getInfoFromBackend();
+
+  readonly menus$ = this.info$.pipe(map((value) => this.getMenu(value || '')));
```
tomalaforge, line 56: you can now use the new signal API but the solution is correct 👍

PR #317, opened 2023-11-16
Answer: 39

Clarchik:


apps/angular/injection-token/src/app/data.ts
```diff
@@ -1 +1,10 @@
-export const DEFAULT_TIMER = 1000;
+import { InjectionToken, Provider } from '@angular/core';
+
+export const TIMER = new InjectionToken<number>('Timer delay');
+
+export function provideTimer(useValue = 1000): Provider {
+  return {
+    provide: TIMER,
+    useValue,
+  };
```
tomalaforge, line 9: 👍

apps/angular/injection-token/src/app/video.component.ts
```diff
@@ -1,10 +1,12 @@
 import { Component } from '@angular/core';
 import { TimerContainerComponent } from './timer-container.component';
+import { provideTimer } from './data';
 
 @Component({
   selector: 'app-video',
   standalone: true,
   imports: [TimerContainerComponent],
+  providers: [provideTimer()],
```
tomalaforge, line 9: you can use a default value inside your injection token to avoid this provider

PR #318, opened 2023-11-16
Answer:3 - ngFor enhancement

DragosAGrigore:


apps/angular/ngfor-enhancement/src/app/ng-for-empty.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: '[ngFor]',
+  standalone: true,
+  hostDirectives: [
+    {
+      directive: NgFor,
+      inputs: ['ngForOf'],
+    },
+  ],
```
tomalaforge, line 20: 👍

apps/angular/ngfor-enhancement/src/app/ng-for-empty.directive.ts
```diff
@@ -0,0 +1,33 @@
...
+export class NgForEmptyDirective<T> implements DoCheck {
+  @Input() ngForOf: T[] = [];
+  @Input() ngForEmpty!: TemplateRef<EmptyListDirective>;
+
+  constructor(private vcr: ViewContainerRef) {}
+
+  ngDoCheck(): void {
+    if (!this.ngForOf.length) {
+      this.vcr.createEmbeddedView(this.ngForEmpty);
+    }
```
tomalaforge, line 31: I didn't test it, but I'm not sure it will work if your list is dynamic. Meaning, if you have element on delete them, I don't think you will show the empty template

jdegand: I added a delete button with a delete method that filtered the this.persons array to exclude a passed in id (added an id to Person object) and the empty prompt showed.

tomalaforge: Ok thanks for testing it.

PR #326, opened 2023-11-17
Answer:17

BruceArmstrong007:

Added Cypress Component and TestingLibrary Unit Test cases


apps/testing/router-outlet/src/app/app.component.spec.ts
```diff
@@ -1,31 +1,98 @@
...
+  };
   describe('Given no search criteria', () => {
     it('Then shows error message and disabled button', async () => {
       //todo
+      await setup();
+      const input = screen.getByRole('textbox');
+      await userEvent.clear(input);
+      expect(
+        await screen.queryByText('Search criteria is required!')
+      ).toBeInTheDocument();
```
tomalaforge, line 27: you don't need to add the expect toBeInTheDocument because the `get`method will already assert that the element you are looking for is inside the document, otherwise it wil throw an error.

BruceArmstrong007: i see, thanks for , i will use
await screen.queryByText('Search criteria is required!')
instead of  expect(
        await screen.queryByText('Search criteria is required!')
      ).toBeInTheDocument();

PR #328, opened 2023-11-17
Answer:18

BruceArmstrong007:

Added testcases with cypress and testing library


apps/testing/nested/src/app/child.component.cy.ts
```diff
@@ -1,18 +1,35 @@
...
   });
 
   describe('When typing "Good" and clicking on Validate', () => {
     test('Then show "Title is Good" message, no error message and send a http request to the backend', async () => {
       setup();
+      cy.get('input[type=text]').clear();
+      cy.get('input[type=text]').type('Good');
+      cy.get('button').click();
+      cy.contains('Title is required !!!').should('not.exist');
+      cy.get('@http').should('not.called');
```
tomalaforge, line 31: are you sure ? Http should be called

apps/testing/nested/src/app/child.component.spec.ts
```diff
@@ -1,16 +1,50 @@
...
+import { createMockWithValues } from '@testing-library/angular/jest-utils';
+import { HttpService } from './http.service';
 
 describe('ChildComponent', () => {
+  const setup = async () => {
+    const httpServiceMock = createMockWithValues(HttpService, {
+      sendTitle: jest.fn(),
+    });
+    await render(ChildComponent, {
+      componentProviders: [
```
tomalaforge, line 13: it should be on the provider property

componentProviders is only to override service provider at component level

apps/testing/nested/src/app/child.component.spec.ts
```diff
@@ -1,16 +1,50 @@
...
+        {
+          provide: HttpService,
+          useValue: httpServiceMock,
```
tomalaforge, line 16: it's nicer this way, why didn't you do it the same way with Cypress ?

apps/testing/nested/src/app/child.component.spec.ts
BruceArmstrong007: i have changed this to providers

PR #334, opened 2023-11-18
Answer:14

BruceArmstrong007:

fixed race condition in component.ts with attaching topicfetchObservable with asyncPipe and ngIf in tempate instead of subscribing in the component
 cypress test case passes


apps/rxjs/race-condition/src/app/app.component.ts
```diff
@@ -2,36 +2,32 @@ import {
...
+  </ng-container> `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class AppComponent implements OnInit {
+export class AppComponent {
   title = 'rxjs-race-condition';
   dialog = inject(MatDialog);
   topicService = inject(TopicService);
-  topics: TopicType[] = [];
+  topics$ = this.topicService.fakeGetHttpTopic().pipe(take(1));
```
tomalaforge, line 25: Perfect, declarative is always better than imperative

PR #338, opened 2023-11-18
Answer:9

AshishKumar409:


apps/angular/pipe-intermediate/src/app/app.component.ts
```diff
@@ -1,34 +1,26 @@
...
 @Component({
   standalone: true,
-  imports: [NgFor],
+  imports: [NgFor, VerifyPipe],
   selector: 'app-root',
   template: `
     <div *ngFor="let person of persons; let index = index; let isFirst = first">
-      {{ showName(person.name, index) }}
-      {{ isAllowed(person.age, isFirst) }}
+      {{ person | verify : index : isFirst }}
```
tomalaforge, line 16: You didn't understand the challenge. You created a specific pipe for this specific use case. The goal of this challenge is to create a **generic** pipe that will work for any function and which is strongly typed

AshishKumar409: got it @tomalaforge, will try to understand it properly.

PR #348, opened 2023-11-21
Answer:14 - Race Condition

DrewOliv:
I arrived at this solution; however, I have doubts about the feasibility of directly using an 'httpClient,' as in the case of 'fakeGetHttpTopic,' within a signal. I believe I may encounter issues, especially in situations that require error handling. Do you agree?

```typescript
topics = toSignal(this.topicService.fakeGetHttpTopic().pipe(take(1)), {
  initialValue: []
});
 ```

### DIFF
--- a/apps/rxjs/race-condition/src/app/app.component.ts
+++ b/apps/rxjs/race-condition/src/app/app.component.ts
@@ -1,37 +1,42 @@
 import {
   ChangeDetectionStrategy,
   Component,
+  computed,
   inject,
-  OnInit,
 } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { MatDialog } from '@angular/material/dialog';
 import { take } from 'rxjs';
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: apps/rxjs/race-condition/src/app/app.component.ts [hunk 1]
```diff
@@ -1,37 +1,42 @@
...
+    initialValue: [],
+  });
 
-  ngOnInit(): void {
-    this.topicService
-      .fakeGetHttpTopic()
-      .pipe(take(1))
-      .subscribe((topics) => (this.topics = topics));
-  }
+  isLoading$ = computed(() => this.topics().length > 0);
```
tomalaforge, line 34: That's not a good condition for the loading indicator. What is topic length is zero ?

tomalaforge:
In case of error handling, if the observable throw an error, `toSignal` will throw an error as well. The best way is to handle that error inside a `catchError` operator and return the shape you want it.

PR #351, opened 2023-11-21
Answer: 16

Clarchik:


apps/angular/di/src/app/currency.directive.ts
```diff
@@ -0,0 +1,15 @@
...
+import { CurrencyService } from './currency.service';
+
+@Directive({
+  selector: '[appCurrency]',
+  standalone: true,
+  providers: [CurrencyService],
+})
+export class CurrencyDirective {
+  private readonly currencySerivce = inject(CurrencyService);
+  @Input('appCurrency') set appCurrency(value: string) {
```
tomalaforge, line 12: if the name of the input is the same, you don't need to remap

tomalaforge:
Perfect great job

PR #354, opened 2023-11-22
Feature/copy command to clipboard

Sagardevkota:
## Add Copy Button for "npx nx ...." Command

- When navigating the Angular Challenges website's documentation, I often find myself manually copying the "npx nx ..." command from the Note section. To enhance user experience, I propose the addition of a convenient copy button. This button will automatically copy the command to the clipboard, allowing users to effortlessly paste it into their terminal for execution.
- New changes would look like following screenshot:

image


tomalaforge:
This is such a nice idea and great improvement. 😍

One small change, can you create a component so we can maybe use it in other places.

tomalaforge:
@all-contributors add @Sagardevkota for code

Sagardevkota:
> This is such a nice idea and great improvement. 😍
> 
> One small change, can you create a component so we can maybe use it in other places. And can you do something like in Github, with a checkmark when the line is copied ?

Thanks for your valuable feedback.. I have created a ClipboardCopy.astro component & moved all the code there.. i am fairly new to astro.. You can point if there is a better way :) .. I think the Github is using setTimeout but i have opted to use mouseleave event.. i can change to whatever you prefer.. 

Thanks Again

tomalaforge:
Nice, better with the component. 

I think I prefer the setTimeout. With mouseLeave, the UX is weird. We don't really understand if the click worked or not. The button is very small, so we move outside of it really fast and the icon reset too fast. A timeout of 3/4s will feel more natural

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,92 @@
...
+
+  let timeout;
+  commandCopyBtn.addEventListener('click', () => {
+    navigator.clipboard.writeText(copyText);
+    tooltipText.innerHTML = 'Copied';
+    copiedCheckMark.style.display = 'inline-block';
+    commandCopyBtn.style.display = 'none';
+    //clear previous timeout
+    clearTimeout(timeout);
+    timeout = window.setTimeout(()=>{
```
tomalaforge, line 51: You don't need to use window.setTimeout, 
you can use setTimeout and don't need to store it and clear it.

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,92 @@
...
+    copiedCheckMark.style.display = 'none';
+    commandCopyBtn.style.display = 'inline-block';
+    tooltipText.innerHTML = 'Copy';
+  }, 750)
```
tomalaforge, line 55: 2000 would be a bit better

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,92 @@
...
+  });
+</script>
+
+<style>
+  .copy-button {
+    fill: var(--sl-color-accent-high);
```
tomalaforge, line 61: I think it's better if it's gray, and red on hover

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,92 @@
...
+    margin-left: 0.5rem;
+    cursor: pointer;
+  }
+
+  .copied-checkmark {
+    fill: var(--sl-color-accent-high);
```
tomalaforge, line 67: like github, green is a good color to show that it worked

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,92 @@
...
+  const commandCopyBtn = document.querySelector('.copy-button');
+  const copiedCheckMark = document.querySelector('.copied-checkmark');
+  const tooltipText = document.querySelector('.tooltipText');
+
+  let timeout;
+  commandCopyBtn.addEventListener('click', () => {
+    navigator.clipboard.writeText(copyText);
+    tooltipText.innerHTML = 'Copied';
+    copiedCheckMark.style.display = 'inline-block';
+    commandCopyBtn.style.display = 'none';
```
tomalaforge, line 48: you can imitate github and show the Copied tooltip when clicked. 👍

tomalaforge:
Very nice, I added a few last comment to make it perfect.

docs/src/components/ClipboardCopy.astro
Sagardevkota: > you can imitate github and show the Copied tooltip when clicked. 👍

Thanks for your feedback. I have changed all the color based on your recommendation & tried to imitate github :smiley

jdegand:
Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL error in console. 

This error is not present on main Angular Challenges app.   

Styling issues : 

Note and Reminder are gone.  

.starlight-aside--note in custom-css file -> need to remove background:transparent.    

The copied text isn't displaying.

Need remove setting the color on .tooltipText (line 82).

Probably best to create a command component with the copy button and add the component where all commands are used in the whole docs ?

Sagardevkota:
> Uncaught (in promise) TypeError: Failed to construct 'URL': Invalid URL error in console.
> 
> This error is not present on main Angular Challenges app.
> 
> Styling issues :
> 
> Note and Reminder are gone.
> 
> .starlight-aside--note in custom-css file -> need to remove background:transparent.
> 
> The copied text isn't displaying.
> 
> Need remove setting the color on .tooltipText (line 82).
> 
> Probably best to create a command component with the copy button and add the component where all commands are used in the whole docs ?

I don't have any errors/issues in my local setup.. I have changed tooltip background to red to fix white bg & color on light theme

jdegand:
Maybe error is caused by problem with vercel deployment.

tomalaforge:
That's perfect, thanks a lot @Sagardevkota

docs/src/components/ClipboardCopy.astro
```diff
@@ -0,0 +1,101 @@
...
+        tooltipText.style.display = 'none';
+  }, 2000)
+  });
+</script>
+
+<style>
+  .copy-button {
+    margin-left: 0.5rem;
+    cursor: pointer;
+  }
```
tomalaforge, line 61: just for information: you can do 
```css
.copy-button {
margin-left: 0.5rem;
    cursor: pointer;
    
    &:hover {
fill: var(--sl-color-accent-high);.}
}
```

tomalaforge (approved):

PR #357, opened 2023-11-23
fix: various typos

jdegand:
I did a little rewording on the pipes section.  

I changed exercice to exercise at one point.   Both are valid.  I reverted it back to stay consistent.


tomalaforge (approved):

tomalaforge:
Thanks a lot 🙏

PR #363, opened 2023-11-24
fix: note & reminder visible on light mode

jdegand:
Had trouble getting astro's preview command to work.

There was an typescript config strict error.  

The note and reminder are not visible on light mode.  I checked the preview seems like nothing else was broken deleting that css.

### DIFF
--- a/docs/src/styles/custom-css.css
+++ b/docs/src/styles/custom-css.css
@@ -139,6 +139,11 @@ starlight-menu-button svg {
   color: var(--sl-color-text);
 }
 
+.starlight-aside--note .starlight-aside__title,
+.starlight-aside--tip .starlight-aside__title {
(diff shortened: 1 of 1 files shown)


docs/src/styles/custom-css.css
```diff
@@ -139,11 +139,6 @@ starlight-menu-button svg {
   color: var(--sl-color-text);
 }
 
-.starlight-aside--note .starlight-aside__title,
-.starlight-aside--tip .starlight-aside__title {
-  color: white;
-}
```
tomalaforge, line 145: no you still need that, otherwise the title is blue: 
but you are right about the light mode, I always forgot to test it. 
you can use 
```css
  color: var(--sl-color-white);
```

tomalaforge (approved):

PR #364, opened 2023-11-24
fix: more typos

jdegand:
There was a typo for styling - write instead of white.  

Sidenote: I think that styling is still overwritten so the sponsor text is still not white on light mode.


tomalaforge (approved):

tomalaforge:
Thanks a lot for all your fixes

PR #368, opened 2023-11-25
feat(Answer:16): solution `Master Dependancy Injection`

volvachev:


apps/angular/di/src/app/app.component.ts
```diff
@@ -37,7 +56,7 @@ export class ProductDirective {
         </tr>
       </ng-template>
       <ng-template #body product let-product>
-        <tr>
+        <tr [currencyCode]="product.currencyCode">
```
tomalaforge, line 59: 👍

PR #373, opened 2023-11-27
Answer:1 projection

stillst:


apps/angular/projection/src/app/ui/card/card-item-tmp.directive.ts
```diff
@@ -0,0 +1,7 @@
+import { Directive } from '@angular/core';
+
+@Directive({
+  standalone: true,
+  selector: 'ng-template[cardItemTmp]',
+})
+export class CardItemTmpDirective {}
```
tomalaforge, line 7: 🔥

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,21 @@
...
-    *ngIf="type === CardType.TEACHER"
-    src="assets/img/teacher.png"
-    width="200px" />
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
-    width="200px" />
-
+  [ngStyle]="{ 'background-color': color }">
+  <ng-content select="[cardImg]" />
```
tomalaforge, line 4: 👍

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,21 @@
...
   <section>
-    <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname"
-      [id]="item.id"
-      [type]="type">
-    </app-list-item>
+    <ng-container *ngFor="let item of list; trackBy: itemById">
```
tomalaforge, line 6: you can now use the new control flow

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,21 @@
...
+      <ng-template
+        *ngTemplateOutlet="
+          cardItemTmp ?? defaultTmp;
+          context: { $implicit: item }
+        " />
```
tomalaforge, line 11: 👍

apps/angular/projection/src/app/app.facade.ts
```diff
@@ -0,0 +1,84 @@
...
+import { CityStore } from './data-access/city.store';
+import { Teacher } from './model/teacher.model';
+import { Student } from './model/student.model';
+import { City } from './model/city.model';
+
+@Injectable()
+export class AppFacade implements OnDestroy {
+  readonly teachers$: Observable<Teacher[]> = this.teacherStore.teachers$;
+  readonly students$: Observable<Student[]> = this.studentStore.students$;
+  readonly cities$: Observable<City[]> = this.cityStore.cities$;
```
tomalaforge, line 21: I don't really understand why you put everything inside the `AppComponent`. The component is handling a lot of logic which should be handle by their own components. 

What will happen if you add one of two more cards, this component will become bigger and bigger. 

What do syou think ?

apps/angular/projection/src/app/ui/card/card.component.html
stillst: I'm still not used to it :)

apps/angular/projection/src/app/app.facade.ts
tomalaforge: Yes because this is a very small application to only work on specific use case. But I get your point, however I'm not very fan of this big app component. I prefer to split the logic per component since each card correspond to a different business domain. So each card item is a smart computer, the `CardComponent` is the dumb component.

stillst: I see, thank you for your feedback!

PR #377, opened 2023-11-29
fix: missing dash in css variable 

jdegand:
Github success btn is missing a dash.  

There are 2 other css problems.  

The sponsor button color in light mode and the color of reactions in dark mode.  Yet to solve either.

### DIFF
--- a/docs/src/styles/custom-css.css
+++ b/docs/src/styles/custom-css.css
@@ -34,7 +34,7 @@
   border-radius: 6px;
   padding: 2px 8px;
   background-color: rgb(35, 134, 54);
-  color: var(-color-btn);
+  color: var(--color-btn);
 }
 
 .github-neutral-btn {


tomalaforge (approved):

tomalaforge:
What do you mean with the reaction button in dark mode ?

tomalaforge:
Ok I see, this will be hard to correct. 🧐

jdegand:
I think the dark mode problems may stem from the binding of the theme.  [See this giscus issue](https://github.com/giscus/giscus/issues/1135).  On my computer the comment section is still white on the dark mode as well.   You are binding in the comment section.astro file with `data-theme="preferred_color_scheme"`.  I noticed inspecting source code both light and dark in some classes.   

Giscus doesn't really seem to support / advertise  Angular support (Giscus component has no version for Angular) and I think the binding issue is probably why.  

It is just a minor issue.

PR #386, opened 2023-12-01
Answer:39

lashagelashvili:


apps/angular/injection-token/src/app/timer.provide.ts
```diff
@@ -0,0 +1,8 @@
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+export const TIMER_TOKEN = new InjectionToken<number>('timer');
+
+export const getTimerProvider = (timer: number): ValueProvider => ({
+  provide: TIMER_TOKEN,
+  useValue: timer,
+});
```
tomalaforge, line 8: I like creating a fonction to provide a TOKEN, but you can use the factory function of the InjectionToken to provide a default value.

alcaidio: he can have a default value with this function 
 const getTimerProvider = (timer = 1000) => {...}
(I do the same of @lashagelashvili)

PR #387, opened 2023-12-02
Answer:5 solution crud tanstack

tomalaforge:


arnoud-dv:
Love it!

It's so easy to read and understand this code and I don't think I can think of any improvements. I did check your other solutions and not having worked with RX myself I found that one especially interesting. But TanStack Query is much simpler to use.

The query status fits very well with the new control flow. I mostly used promises so far when experimenting with Angular Query. Would like to add first-class support for observables but using observables returned from a service and converting them in the queryFn / mutationFn looks much better than I expected. I think it's a better approach than converting it to a promise in the service.

In a larger application I like the idea of using a [key factory](https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories)

There is also [this library](https://github.com/lukemorales/query-key-factory) linked in the React Query documentation I find really interesting. It includes the queryFn in the key factory. It does have a dependency on React Query, but would probably be easy to convert to Angular Query.

Thanks, I've mostly been working on the adapter itself and seeing people building stuff with it is really cool!

tomalaforge:
thanks a lot for your feedback. I thought of doing a factory for the keys, but I should at least create a constant for this simple use case to avoid magic string. 

I just find that having two mutations inside the same component make the condition not very simple. 

And the mutationFn returning a promise is great and easy to adapt if we want to use axios or fetch. 

The library is very nice and easy to understand, and compare to componentStore or rxAngular, it's way simpler to create proper code. 

Great job on doing the adapter. 👍 I need to buy a more complex use case to really get the full power of it.

PR #392, opened 2023-12-06
[Answer:1]: Solution projection with Signals and new control flow

blugrinc:

- Refactored the `CardComponent` and `ListItemComponent`
- Kept `NgFor `directive inside the `CardComponent`
- `CardComponent `does not contain any `NgIf `or `NgSwitch`
- Removed `::ng-deep` from component styles


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,34 +1,66 @@
...
+  constructor(public templateRef: TemplateRef<unknown>) {}
+}
+
 @Component({
   selector: 'app-card',
-  templateUrl: './card.component.html',
+  template: ` <ng-content select="[img-header]"></ng-content>
+
+    <section>
+      @if (list.length > 0) {
```
tomalaforge, line 27: @for has an empty block.

apps/angular/projection/src/app/data-access/city.store.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject } from 'rxjs';
+import { City } from '../model/city.model';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class CityStore {
+  private citys = new BehaviorSubject<City[]>([]);
+  citys$ = this.citys.asObservable();
```
tomalaforge, line 10: you can use a signal here instead, no ? signal are good for state management

PR #393, opened 2023-12-07
Answer: 1 Projection challenge

sidarth-23:
## My questions are

1. Advantages of signals over behaviour subject as I'm not able to find good examples on that ?
2. Advantages of ng-content as the new control flow is seeming to take of this well ?


package.json
```diff
@@ -41,7 +41,7 @@
     "zone.js": "0.14.2"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "17.0.1",
+    "@angular-devkit/build-angular": "^17.0.6",
```
tomalaforge, line 44: Why did you change the version ?

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,23 @@
...
-    src="assets/img/student.webp"
+    [src]="imageUrl"
     width="200px" />
 
   <section>
+    @for (item of list; track $index) {
     <app-list-item
-      *ngFor="let item of list"
-      [name]="item.firstname"
+      [name]="item.name"
```
tomalaforge, line 12: This is not very customizable. What will you do if you need to add an icon but only for one card, or a new button ? Look at ngTemplateOutlet

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -1,27 +1,23 @@
...
-  [class]="customClass">
+  [class]="color">
   <img
-    *ngIf="type === CardType.TEACHER"
-    src="assets/img/teacher.png"
-    width="200px" />
-  <img
-    *ngIf="type === CardType.STUDENT"
-    src="assets/img/student.webp"
+    [src]="imageUrl"
```
tomalaforge, line 6: prefer content projection using ng-content over input. More flexible and customizable

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -12,23 +18,37 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
 
+  ngOnInit(): void {
+    if (this.type === CardType.TEACHER) {
+      this.imageUrl = 'assets/img/teacher.png';
+    } else if (this.type === CardType.STUDENT) {
+      this.imageUrl = 'assets/img/student.png';
+    } else if (this.type === CardType.CITY) {
+      this.imageUrl = 'assets/img/city.png';
+    }
+  }
```
tomalaforge, line 43: this component is meant to be dumb. No logic should stay here. And the list will grow if you add more cardType. Look at @Output

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,43 @@
...
 
-  ngOnInit(): void {}
+  constructor(
+    private http: FakeHttpService,
+    private store: CityStore,
+  ) {}
+
+  ngOnInit(): void {
+    this.http.fetchCities$.subscribe((c) => this.store.addAll(c));
+    this.store.city$.subscribe((c) => {
```
tomalaforge, line 32: avoid imperative coding, prefer declarative using Rxjs or Signals. (But prefer Signals when it comes to template binding)

tomalaforge:
Thanks for submitting your solution. There is a lot of work to be done. You didn't handle that `CardComponent` is a UI component. No logic and no if or condition should be handle in that component. If you have any question, feel free to ask them here.

tomalaforge:
For the two questions: Signals is very similar to behavior subject in how they behave. However it's not the same reactive pattern. Signals is very good for any synchronous logic, store management, template binding, etc. Rxjs is better for event, http request. 

I think most of the time you should go for signals. Angular will be using more and more signals in the future. 

2. I didn't get your quesiton. Control flow and ng-content are two separates things. Control flow is to handle for loop, condition, ng-content is for projection content into a child

sidarth-23:
> For the two questions: Signals is very similar to behavior subject in how they behave. However it's not the same reactive pattern. Signals is very good for any synchronous logic, store management, template binding, etc. Rxjs is better for event, http request.
> 
> I think most of the time you should go for signals. Angular will be using more and more signals in the future.
> 
> 2. I didn't get your quesiton. Control flow and ng-content are two separates things. Control flow is to handle for loop, condition, ng-content is for projection content into a child

I feel that both using ng-content and @if gives me the same output at the end. Can i know in specific what makes the 2 different?

tomalaforge:
ng-content will project content from the parent to the child

@if will display the html inside the condition if the condition is met. 

This is completely different. But of course, you can input a condition and display the same content and it will give you the same result, but the logic will be handle inside the child and not the parent which make the UI component not very customizable

sidarth-23:
> ng-content will project content from the parent to the child
> 
> @if will display the html inside the condition if the condition is met. 
> 
> This is completely different. But of course, you can input a condition and display the same content and it will give you the same result, but the logic will be handle inside the child and not the parent which make the UI component not very customizable

I'm starting to get a picture of it now. Thanks for the nice explanation. 

I'll refactor my code now and create a pr again

tomalaforge:
Invalid

PR #401, opened 2023-12-11
Answer 8: pure pipe challenge

sidarth-23:

- Added custom pipe to handle the array

I think this is what the challenge asked for. What is the more modular way to approach the problem?

### DIFF
--- a/apps/angular/pipe-easy/src/app/app.component.ts
+++ b/apps/angular/pipe-easy/src/app/app.component.ts
@@ -1,21 +1,19 @@
 import { NgFor } from '@angular/common';
 import { Component } from '@angular/core';
+import { ArrayToTextPipe } from './array-to-text.pipe';
 
 @Component({
   standalone: true,
-  imports: [NgFor],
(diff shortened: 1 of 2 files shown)


apps/angular/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,19 @@
...
   standalone: true,
-  imports: [NgFor],
+  imports: [NgFor, ArrayToTextPipe],
   selector: 'app-root',
   template: `
-    <div *ngFor="let person of persons; let index = index">
-      {{ heavyComputation(person, index) }}
+  @for (person of persons; let index = $index ;track person) {
+    <div>
+      {{ person | arrayToText: index }}
```
tomalaforge, line 12: That's perfect, you can even call `$index` immediately without creating an alias

sidarth-23: Oh that's nice. Thanks buddy

PR #403, opened 2023-12-12
docs: clarify command

jdegand:
May be best to add a hint that you need to use npx to generate the web worker.  

Sidenotes:  

- You are missing a link on line 22. 

- `First, create a web worker using the schematics, then move the issuing function.`  

If you pick the regular angular schematic, the web worker conditional boilerplate is not added to the app component.   The conditional is added when you use the nx schematic.   Add as hint to the readme to pick the nx schematic ?

Are you supposed to have the progress percentage update or not?

### DIFF
--- a/docs/src/content/docs/challenges/performance/40-christmas-web-worker.md
+++ b/docs/src/content/docs/challenges/performance/40-christmas-web-worker.md
@@ -29,4 +29,4 @@ First, create a web worker using the schematics, then move the issuing function.
 
 ### Note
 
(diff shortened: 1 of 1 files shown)


tomalaforge:
Thanks for your feedback, what do you think of the new change ?

jdegand:
Much better.

tomalaforge:
@jdegand Why did you close it ?

tomalaforge (approved):

PR #407, opened 2023-12-12
Answer 40: web worker

jdegand:
Great challenge.

I struggled and tried a few different approaches.  

Solution could probably be improved.  I need to look into error handling (`worker.onerror`).

### DIFF
--- a/apps/performance/christmas-web-worker/src/app/app.component.ts
+++ b/apps/performance/christmas-web-worker/src/app/app.component.ts
@@ -16,7 +16,9 @@ import { UnknownPersonComponent } from './unknown-person/unknown-person.componen
       (click)="discover()">
       Discover
     </button>
-    <div class="p-1 text-white">Progress: {{ loadingPercentage }}%</div>
+    <div class="p-1 text-white">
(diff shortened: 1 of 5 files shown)


apps/performance/christmas-web-worker/src/app/app.component.ts
```diff
@@ -24,9 +27,17 @@ import { UnknownPersonComponent } from './unknown-person/unknown-person.componen
...
+  loadingPercentage = 0;
 
   discover() {
-    this.heavyCalculationService.startLoading();
+    if (typeof Worker !== 'undefined') {
+      const worker = new Worker(new URL('./app.worker', import.meta.url));
+      worker.onmessage = ({ data }) => {
+        this.loadingPercentage = data;
+      };
+      worker.postMessage('start'); // what you type here doesn't matter -> like a key
```
tomalaforge, line 38: That's true, in our case, the message doesn't matter

tomalaforge:
Very nice work. I also struggle a bit on the communication between the worker and the main thread.

PR #419, opened 2023-12-17
Answer:1  Refactor card and list-item to use Content Projection

grilled-pasta:

## Checklist for Constraints
- [x] You must refactor the `CardComponent` and `ListItemComponent`.
- [x] The `NgFor` directive must be declared and remain inside the `CardComponent`. You might be tempted to move it to the `ParentCardComponent` like `TeacherCardComponent`.
- [x] `CardComponent` should not contain any `NgIf` or `NgSwitch`.
- [x] `CSS`: try to avoid using `::ng-deep`. Find a better way to handle `CSS` styling.

## Checklist for Bonus Challenges

- [x] Try to work with the new built-in control flow syntax for loops and conditionals
- [x] Use the signal API to manage your components' state 
- [x] To reference the template, use a directive instead of magic strings


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,72 @@
...
-        Add
-      </button>
-    </div>
+    <ng-content select="[card-img]"></ng-content>
+
+    <section>
+      @for (item of list; track item.id) {
+        <ng-container
+          [ngTemplateOutlet]="listItemTemplate.templateRef"
+          [ngTemplateOutletContext]="{ $implicit: item }"></ng-container>
```
tomalaforge, line 37: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,72 @@
...
-          [type]="type"></app-list-item>
-      </section>
-
-      <button
-        class="rounded-sm border border-blue-500 bg-blue-300 p-2"
-        (click)="addNewItem()">
-        Add
-      </button>
-    </div>
+    <ng-content select="[card-img]"></ng-content>
```
tomalaforge, line 31: nit: you can even write `<ng-content select="[card-img]" />`

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,72 @@
...
-      this.studentStore.addOne(randStudent());
-    }
-  }
+export class CardComponent<T extends CardItem> {
+  @Input() list!: T[];
+
+  @Output() add = new EventEmitter<void>();
+
+  @ContentChild(CardContentDirective)
+  listItemTemplate!: CardContentDirective<T>;
```
tomalaforge, line 64: 👍

tomalaforge:
Very nice work and I very like your `item.store` to factorize all store. Very great job

grilled-pasta:
Thank you! I had trouble understanding content projection with templates but after this challenge I can refactor my own projects. 

I'm not sure if it's a good practice to use the 'CardItem' interface for both the store and the UI components. Do you have any suggestions on best practices for such cases?

tomalaforge:
I think for your second question it's only because the interface is poorly named. It doesn't really reflect the reality. You are only extending your store or cardItem to validate that your object has an id. Maybe an other name can be a better choice. 

I rarely create an interface for id only. I always prefer to write `<T extends {id: string}>` which is more readable and you get the code immediatly.

PR #424, opened 2023-12-18
chore: run prettier on files in control-value-accessor folder

Pascalmh:
The diff might be bigger that it needs to be for Challenge 41. 

I ran prettier on all files in the `apps/forms/control-value-accessor` folder to make the diff of submitted answers easier to review.


tomalaforge:
Nice thanks for that 👍

PR #434, opened 2023-12-21
fix: change placeholder text

jdegand:
The placeholder text 'write an article' makes no sense.


tomalaforge (approved):

tomalaforge:
😅 That's a detail. but you are right 👍

PR #438, opened 2023-12-22
Answer:1

nutphi:

- [x] Start your PR message with Answer:1

## Constraints

- [x] You must refactor the CardComponent and ListItemComponent.
- [x] The NgFor directive must be declared and remain inside the CardComponent. You might be tempted to move it to the ParentCardComponent like TeacherCardComponent.
- [x] CardComponent should not contain any NgIf or NgSwitch.
- [x] CSS: try to avoid using ::ng-deep. Find a better way to handle CSS styling.

## Bonus Challenges
- [x] Try to work with the new built-in control flow syntax for loops and conditionals (documentation [here](https://angular.dev/guide/templates/control-flow))
- [x] Use the signal API to manage your components state (documentation [here](https://angular.dev/guide/signals))
- [x] To reference the template, use a directive instead of magic strings

### DIFF
--- a/apps/angular/projection/src/app/data-access/city.store.ts
+++ b/apps/angular/projection/src/app/data-access/city.store.ts
@@ -1,23 +1,15 @@
-import { Injectable } from '@angular/core';
-import { BehaviorSubject } from 'rxjs';
+import { Injectable, effect } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { City } from '../model/city.model';
+import { ShareStore } from './share.store';
 
 @Injectable({
   providedIn: 'root',
 })
-export class CityStore {
-  private cities = new BehaviorSubject<City[]>([]);
-  cities$ = this.cities.asObservable();
-
-  addAll(cities: City[]) {
-    this.cities.next(cities);
-  }
-
-  addOne(student: City) {
-    this.cities.next([...this.cities.value, student]);
-  }
-
-  deleteOne(id: number) {
-    this.cities.next(this.cities.value.filter((s) => s.id !== id));
+export class CityStore extends ShareStore<City> {
+  constructor() {
+    super();
+    const _cities = toSignal(this.http.fetchCities$, { initialValue: [] });
+    effect(() => this.addAll(_cities()), { allowSignalWrites: true });
   }
 }
--- a/apps/angular/projection/src/app/ui/card/card.component.ts
+++ b/apps/angular/projection/src/app/ui/card/card.component.ts
@@ -1,61 +1,62 @@
-import { NgFor, NgIf } from '@angular/common';
-import { Component, Input } from '@angular/core';
-import { randStudent, randTeacher } from '../../data-access/fake-http.service';
-import { StudentStore } from '../../data-access/student.store';
-import { TeacherStore } from '../../data-access/teacher.store';
-import { CardType } from '../../model/card.model';
-import { ListItemComponent } from '../list-item/list-item.component';
+import { NgTemplateOutlet } from '@angular/common';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  ContentChild,
+  EventEmitter,
+  Input,
+  Output,
+  TemplateRef,
+} from '@angular/core';
+import { ListItemTemplateDirective } from '../../list-item-template.directive';
+import { CardItem } from '../../model/card.model';
 
 @Component({
   selector: 'app-card',
   template: `
-    <div
-      class="flex w-fit flex-col gap-3 rounded-md border-2 border-black p-4"
-      [class]="customClass">
-      <img
-        *ngIf="type === CardType.TEACHER"
-        src="assets/img/teacher.png"
-        width="200px" />
-      <img
-        *ngIf="type === CardType.STUDENT"
-        src="assets/img/student.webp"
-        width="200px" />
+    <div class="flex w-fit flex-col gap-3 rounded-md border-2 border-black p-4">
+      <ng-content select="[img]"></ng-content>
 
       <section>
-        <app-list-item
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
+          <ng-container
+            *ngTemplateOutlet="
+              template;
+              context: { $implicit: item }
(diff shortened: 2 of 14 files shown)


apps/angular/projection/src/app/data-access/city.store.ts
```diff
@@ -1,23 +1,15 @@
...
-    this.cities.next([...this.cities.value, student]);
-  }
-
-  deleteOne(id: number) {
-    this.cities.next(this.cities.value.filter((s) => s.id !== id));
+export class CityStore extends ShareStore<City> {
+  constructor() {
+    super();
+    const _cities = toSignal(this.http.fetchCities$, { initialValue: [] });
+    effect(() => this.addAll(_cities()), { allowSignalWrites: true });
```
tomalaforge, line 13: I think, it's not a good think to want to do everything with signals. 

Observable or Promise will not disappear, and having a subscribe on the http call is better than having this code. 

This code is hard to understand, less maintainable. 
if you write 

```ts
this.http.fetchCities().pipe(take(1)).subscribe(c => this.addAll(c))
```

This is easier to understand.

What do you think?

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,63 @@
...
       [class]="customClass">
-      <img
-        *ngIf="type === CardType.TEACHER"
-        src="assets/img/teacher.png"
-        width="200px" />
-      <img
-        *ngIf="type === CardType.STUDENT"
-        src="assets/img/student.webp"
-        width="200px" />
+      <ng-content select="[img]"></ng-content>
```
tomalaforge, line 19: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,63 @@
...
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
+          <ng-container
+            *ngTemplateOutlet="
+              template;
+              context: { $implicit: item }
+            "></ng-container>
```
tomalaforge, line 27: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,63 @@
...
         class="rounded-sm border border-blue-500 bg-blue-300 p-2"
-        (click)="addNewItem()">
+        (click)="this.add.emit()">
         Add
       </button>
     </div>
   `,
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
+  imports: [NgIf, NgFor, NgTemplateOutlet],
```
tomalaforge, line 39: NgIf and NgFor should not be here

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,63 @@
...
-  ) {}
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @ContentChild(TemplateRef) template!: TemplateRef<{ $implicit: CardItem }>;
```
tomalaforge, line 61: Use the TemplateRef is dangerous because if you decide to add a new template later, or someone in your team, you can have weird bug or side effect. 

You should prefer to use a directive to target to right template

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,30 @@
...
-  ) {}
-
-  delete(id: number) {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.deleteOne(id);
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.deleteOne(id);
-    }
+  del(id: number) {
+    this.remove.emit(id);
```
tomalaforge, line 28: id is not necessary. This can only be a event trigger

tomalaforge:
Nice work. I added a few comments to improve it even further

apps/angular/projection/src/app/data-access/city.store.ts
nutphi: I think the observable approach is often more familiar to any angular developers.
and the use of effect and toSignal is not that hard to understand I feel it's more understandable than using a lot of observable techniques. and we can avoid the subscription and let angular handle itself with signal approach. for maintainability, you can add/update/debug the code inside effect.

what do you think?

tomalaforge: for me, the code you wrote is mostly a hack of the signal API. Effect should be used as a last resort. Using `allowSignalWrites` is kind of a warning that this is not the way to go. 

I will always use observable or promise for http event. 

But if you feel like you prefer this approach, you can try it on bigger project and see how it goes. 😇

nutphi: yeah, you're right. allowSignalWrite is not recommended. Actually, I feel Signal is still under development by angular team. but I just try to get a better angular 17 signal approach instead of using the old rxjs technique. :)

Thank you for all suggestions.

apps/angular/projection/src/app/ui/card/card.component.ts
nutphi: I have updated it with empty directive. what do you think about it?

```
import { Directive } from '@angular/core';
@Directive({
  selector: '[list-item-template]',
  standalone: true,
})
export class ListItemTemplateDirective {}
```

and add it inside ContentChild directive

```
@ContentChild(ListItemTemplateDirective, { read: TemplateRef })
  template!: TemplateRef<{ $implicit: CardItem }>;
```

apps/angular/projection/src/app/data-access/city.store.ts
tomalaforge: Yes effect is still experimental, but all the rest of the API is stable.

apps/angular/projection/src/app/ui/card/card.component.ts
tomalaforge: yes that's perfect

PR #442, opened 2023-12-26
Answer:4  Typed Context Outlet

jdegand:
I added a lot of comments and did a fair amount of reworking the code.  I didn't use star syntax.     

I used Angular 17 control flow syntax in the list component.

I think adding a screenshot of the desired html result to the directions could be beneficial.


tomalaforge:
very nice 👍

PR #444, opened 2023-12-26
Answer:5 ngrx/signal-store

nutphi:

- [x] Start your PR message with Answer:5

## Step 1: refactor with best practices
What you will need to do:
- [x] Avoid any as a type. Using Interface to leverage Typescript type system prevent errors
- [x] Use a separate service for all your http calls and use a Signal for your todoList
- [x] Don’t mutate data

## Step 2: Improve
- [x] Add a Delete button: [Doc of fake API](https://jsonplaceholder.typicode.com/)
- [x] Handle errors correctly. (Globally)
- [x] Add a Global loading indicator. You can use MatProgressSpinnerModule

## Step 3: Maintainability!! add some test
- [x] Add 2/3 tests
## Step 4: Awesomeness!!! master your state.
- [x] Use the component store of ngrx, ngrx/store, rxAngular, tanstack-query or **ngrx/signal-store** as a local state of your component.
- [x] Have a localized Loading/Error indicator, e.g. only on the Todo being processed and disable all buttons of the processed Todo. (Hint: you will need to create an ItemComponent)

### DIFF
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,51 +1,46 @@
 import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
-import { randText } from '@ngneat/falso';
+import { Component, inject } from '@angular/core';
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { TodosStore } from './global-store';
+import { TodoComponent } from './todo.component';
+import { TodoService } from './todo.service';
 
 @Component({
   standalone: true,
-  imports: [CommonModule],
+  imports: [CommonModule, MatProgressSpinnerModule, TodoComponent],
   selector: 'app-root',
+  styles: `:host {
+    --mdc-circular-progress-active-indicator-color: grey;
+    --mat-icon-color: red;
+    dialog {
+      margin: 0;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+    }
+  }`,
   template: `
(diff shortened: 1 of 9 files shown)


apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,46 @@
...
+      transform: translate(-50%, -50%);
+    }
+  }`,
   template: `
-    <div *ngFor="let todo of todos">
-      {{ todo.title }}
-      <button (click)="update(todo)">Update</button>
-    </div>
+    @if (store.state(); as state) {
+      @if (state == 'loading') {
```
tomalaforge, line 25: nit: always prefer ===

apps/angular/crud/src/app/global-store.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+    deleteItem: (id: number) => void;
+    resetState: () => void;
+    disableButtons: () => void;
+    enableButtons: () => void;
+  } & StateSignal<{
+    state: TodoState;
+    disabled: boolean;
+    entityMap: EntityMap<Todo>;
+    ids: EntityId[];
+  }>;
```
tomalaforge, line 29: why did you need to write this ?

tomalaforge:
Thanks a lot for your answer. I have never played with the signal store yet, so I was very eager to see your solution,

apps/angular/crud/src/app/global-store.spec.ts
nutphi: The first part of the assignment is used for linting.
For the StateSignal, It is actually generated by VSCode, which is not necessary for testing.
Thanks for validation. :)

PR #445, opened 2023-12-26
Answer:8

nutphi:

- [x] Start your PR message with Answer:8


apps/angular/pipe-easy/src/app/heavy-computation.pipe.ts
```diff
@@ -0,0 +1,12 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'heavyComputation',
+  standalone: true,
+  pure: true,
```
tomalaforge, line 6: This is the default setting. You don't have to mention it.

PR #446, opened 2023-12-27
Answer:2

ronald-pacheco-dchs:


apps/ngrx/effect-selector/src/app/store/activity/activity.selectors.ts
```diff
@@ -1,10 +1,30 @@
...
+    if (!user?.isAdmin) return [];
+    return activities.map((a) => {
+      return {
+        ...a,
+        teachers: activities
+          .filter((ac) => ac.type === a.type)
+          .map((ac) => {
+            return { name: ac.teacher.name };
+          }),
+      } as ActivityWithTeachers;
```
tomalaforge, line 27: I better to avoid the `as` keyword. 

Prefer to add a return type instead of as.

tomalaforge:
That a very good work. You can use the new control flow to get use to it otherwise that's perfect.

PR #449, opened 2023-12-27
Answer:31

nutphi:

- [x] Start your PR message with Answer:31


tomalaforge:
module should not exist anymore. That's the goal of this challenge

nutphi:
> module should not exist anymore. That's the goal of this challenge

nutphi:
I'm sorry. I closed the PR by mistake. I've updated the fix and removed all modules.

tomalaforge:
Perfect, that's how we do it. Great work

You can use default route to avoid the `_=>_.xxx` but this a preference.

PR #452, opened 2023-12-27
Answer:35

nutphi:

- [x] Start your PR message with Answer:35

### DIFF
--- a/apps/performance/memoized/src/app/fibonacci.pipe.ts
+++ b/apps/performance/memoized/src/app/fibonacci.pipe.ts
@@ -1,0 +1,31 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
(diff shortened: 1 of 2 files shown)


apps/performance/memoized/src/app/fibonacci.pipe.ts
```diff
@@ -0,0 +1,31 @@
...
+      return 1;
+    }
+    const result = this.getNFibonacci(num - 1) + this.getNFibonacci(num - 2);
+    this.setNFibonacci(num, result);
+    return result;
+  }
+
+  transform(num: number): number {
+    return this.fibonacci(num);
+  }
```
tomalaforge, line 30: nice solution. The pipe is already memoized, but doesn't keep the cache if the page is refreshed, indeed

PR #455, opened 2023-12-28
Answer:14

nutphi:

- [x] Start your PR message with Answer:14

### DIFF
--- a/apps/rxjs/race-condition/src/app/app.component.ts
+++ b/apps/rxjs/race-condition/src/app/app.component.ts
@@ -1,38 +1,22 @@
-import {
-  ChangeDetectionStrategy,
-  Component,
-  inject,
-  OnInit,
-} from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject } from '@angular/core';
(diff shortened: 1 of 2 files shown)


apps/rxjs/race-condition/src/app/app.component.ts
```diff
@@ -12,27 +7,26 @@ import { TopicService, TopicType } from './topic.service';
...
-      .subscribe((topics) => (this.topics = topics));
-  }
-
-  openTopicModal() {
-    this.dialog.open(TopicModalComponent, {
-      data: {
-        topics: this.topics,
-      },
-    });
+      .subscribe((topics: TopicType[]) => {
```
tomalaforge, line 24: why not but it means that you will need to make a http call each time you open the modal. But maybe you want fresh data each time. But in that case, it's maybe better to make the call inside the modal directly. because if the call takes 10s, the modal will not open and the UX is not good for the user.

nutphi: Yeah, you're right. I did not notice it. :) Thanks for validation,

PR #457, opened 2023-12-29
Answer:3

nutphi:

> NOTE: This exercice can feel obsolete with the new control flow and the empty block inside the @for block. However structural directives are not going to be deleted any time soon, so you can still learn a lot from this exercice.
- from the note, I provided the new control flow and the empty block inside the @for block instead :)

### DIFF
--- a/apps/angular/ngfor-enhancement/src/app/app.component.ts
+++ b/apps/angular/ngfor-enhancement/src/app/app.component.ts
@@ -1,4 +1,3 @@
-import { NgFor, NgIf } from '@angular/common';
 import { ChangeDetectionStrategy, Component } from '@angular/core';
 
 interface Person {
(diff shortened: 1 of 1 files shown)


apps/angular/ngfor-enhancement/src/app/app.component.ts
```diff
@@ -7,15 +6,13 @@ interface Person {
...
-      <div *ngFor="let person of persons">
-        {{ person.name }}
-      </div>
-    </ng-container>
-    <ng-template #emptyList>The list is empty !!</ng-template>
+    @for (person of persons; track person.name) {
+      <div>{{ person.name }}</div>
+    } @empty {
+      The list is empty !!
+    }
```
tomalaforge, line 15: ahah, This challenge is becoming obsolete with the new control flow. I should refactor it to enhance another directive. Enhancing directive is still a good knowlegde to have. Structural directive are not going anywhere soon. 

But for this challenge, this solution works

PR #460, opened 2023-12-30
Answer:30 ngrx/signal-store

nutphi:

### DIFF
--- a/apps/angular/interop-rxjs-signal/src/app/list/photos.model.ts
+++ b/apps/angular/interop-rxjs-signal/src/app/list/photos.model.ts
@@ -1,0 +1,22 @@
+import { Photo } from '../photo.model';
+
+export const STATE_KEYS = ['photos', 'total', 'page', 'term'] as const;
+export type StateKey = (typeof STATE_KEYS)[number];
+
+export type State = {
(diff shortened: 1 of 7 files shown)


apps/angular/interop-rxjs-signal/src/app/list/photos.model.ts
```diff
@@ -0,0 +1,22 @@
...
+  [key in StateKey]: key extends 'photos'
+    ? Photo[]
+    : key extends 'total'
+      ? number
+      : key extends 'page'
+        ? number
+        : key extends 'term'
+          ? string
+          : never;
+};
```
tomalaforge, line 16: Why did you create this State type ? 
Why not creating the interface you really need ?

tomalaforge:
Very nice solution with the signal store. I still haven't use it, so it's great to have someone showing how to use it 👍

apps/angular/interop-rxjs-signal/src/app/list/photos.model.ts
nutphi: it is my fault. it is automatically generated by vscode. I've fixed it.

PR #470, opened 2024-01-04
Answer:42 nx static import

jdegand:
This is the original output when you run the app.
old

This is after splitting the library.  
new

I got rid of the original eslint error, but now there is a `no-host-metadata-property` eslint error in the new `user` library. 

The use of crypto causes an error in the browser console.  I don't think it is worth adding `crypto-browserify` for a polyfill.  Of course, that library might not work, as not all crypto functions have polyfills.


tsconfig.base.json
```diff
@@ -77,9 +80,7 @@
       "@tomalaforge/ngrx-callstate-store": [
         "libs/shared/ngrx-callstate-store/src/index.ts"
       ],
-      "@angular-challenges/static-dynamic-import/users": [
-        "libs/static-dynamic-import/users/src/index.ts"
-      ]
+      "user": ["libs/user/src/index.ts"]
```
tomalaforge, line 83: you should be more consistent with you ts link.

libs/static-dynamic-import/users/tsconfig.lib.json
```diff
@@ -8,5 +8,9 @@
     "types": []
   },
   "exclude": ["src/**/*.spec.ts", "jest.config.ts", "src/**/*.test.ts"],
-  "include": ["src/**/*.ts"]
+  "include": [
+    "src/**/*.ts",
+    "../../user/src/lib/user.model.ts",
+    "../../user/src/lib/user.pipe.ts"
```
tomalaforge, line 14: be careful when moving you file, vscode will add the path here

libs/user/src/lib/user.pipe.ts
```diff
@@ -10,3 +10,5 @@ export class UserPipe implements PipeTransform {
     return `${user.name} ${user.lastname} - ${user.country}`;
   }
 }
+
+// https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin/docs/rules/no-host-metadata-property.md
```
tomalaforge, line 14: I don't get it. I don't have any eslint error, did you try to restart your linter ?

tomalaforge:
for better consistency, you could have moved the lib inside `static-dynamic-import` folder, so all libs related to that project go inside this folder.

libs/static-dynamic-import/users/tsconfig.lib.json
jdegand: I didn't want to have to move the files, but the generate command wouldn't execute inside multiple sub-folders.   I don't have `nx` installed globally.   It is good to know that if you move files, you have to do a cleanup afterwards.

libs/user/src/lib/user.pipe.ts
jdegand: Didn't help.  There is probably a misconfiguration somewhere.

libs/static-dynamic-import/users/tsconfig.lib.json
tomalaforge: I don't think you need nx install globally, npx nx or nx console will work if nx is install inside your project.

libs/user/src/lib/user.pipe.ts
tomalaforge: This is weird indeed

libs/static-dynamic-import/users/tsconfig.lib.json
jdegand: The generate command changed from `library` to `lib`.  I think I used `library` and it still worked.

PR #472, opened 2024-01-04
Answer:31 My resolution -> module to standalone

Ruiagf:
Answer:31


tomalaforge:
Nice work but you have a lot of noise with the prettier config and file that are saved but are not inside this challenge scope. 

Personal choice: but I prefer to use the default keywork to avoid the `_ => _.xxx`

Ruiagf:
Hi @tomalaforge!


Thanks for your feeback. : :pray: :slightly_smiling_face:

>  you have a lot of noise with the prettier config and file that are saved but are not inside this challenge scope.

You're right. I will take that in consideration in the following challenges :blush:


Best regards!

PR #473, opened 2024-01-05
Answer:1 angular-projection 

LKL2017:
This is my first step in **angular-challenges** and someting to share below:

1.  I removed the conditional branch and placed it in a appropriate postion.
2.  I'm not sure how to style the element in `ng-content ` well.
3. Does the provider increased the complexity in this case?
4. The computed singal dont have `[Symbol.iterator()]` interface so it cant be used in `*ngFor` or `@for` flow control. Since its a devloper preview feature, I think I could keep following or give some feedback to Angular teams.

---
btw, I haven't code with Angular in a while, and I think this is a pretty good platform to share idea, communicate and sharpen skills in many dimensions. Thanks for offering this excellent🎉 project.

### DIFF
--- a/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
+++ b/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
@@ -9,17 +9,14 @@ import { CardComponent } from '../../ui/card/card.component';
 @Component({
   selector: 'app-teacher-card',
   template: `
-    <app-card [list]="teachers" [type]="cardType" customClass="bg-light-red">
+    <app-card
+      [list]="teachers"
+      [type]="cardType"
+      customClass="bg-red-200/50"
+      showingProp="firstName">
       <img src="assets/img/teacher.png" alt="" />
     </app-card>
   `,
-  styles: [
-    `
-      ::ng-deep .bg-light-red {
-        background-color: rgba(250, 0, 0, 0.1);
-      }
-    `,
-  ],
   standalone: true,
   imports: [CardComponent],
   providers: [{ provide: Store, useExisting: TeacherStore }],
--- a/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
+++ b/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
@@ -1,5 +1,6 @@
 import { Component, OnInit } from '@angular/core';
 import { FakeHttpService } from '../../data-access/fake-http.service';
+import { Store } from '../../data-access/store';
 import { TeacherStore } from '../../data-access/teacher.store';
 import { CardType } from '../../model/card.model';
 import { Teacher } from '../../model/teacher.model';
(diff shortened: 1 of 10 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -37,25 +28,22 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
+  @Input() showingProp!: string;
   @Input() customClass = '';
 
-  CardType = CardType;
-
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore,
-  ) {}
+  constructor(@SkipSelf() private storeService: Store<CardItem>) {}
```
tomalaforge, line 39: you should have no logic inside this component. This component is a presentational component and its only purpose is to display the data. 

Use @Ouput to dispatch event to the parent. 
`ng-content` to project content
and `ngTemplateOulet` to project more complex content.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -12,21 +9,15 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
+          <app-list-item
+            [name]="item[showingProp]"
+            [id]="item.id"
+            (deleteItem)="deleteItem($event)"></app-list-item>
+        }
```
tomalaforge, line 20: this is too specific and not very customizable. If I want to add an icon for only one card, this will be complicated to add.

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,35 +1,38 @@
...
-    `,
-  ],
+  template: `
+    <app-card
+      [list]="teachers"
+      [type]="cardType"
+      customClass="bg-red-200/50"
+      showingProp="firstName">
+      <img src="assets/img/teacher.png" alt="" />
+    </app-card>
```
tomalaforge, line 18: The goal is to have no @Input or almost none

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,35 +1,38 @@
...
+  `,
   standalone: true,
   imports: [CardComponent],
+  providers: [{ provide: Store, useExisting: TeacherStore }],
```
tomalaforge, line 22: Ok I got why you did that. That's an idea, but too complex. Let the logic inside this component, and the card component will only dispatch add and delete action

tomalaforge:
Hello, good work but you are doing thing way too complicated. 
Presentation components handle UI
Logical components handle Logic

LKL2017:
@tomalaforge  Thanks for you attention and reply, maybe I should achieve the goal as a more targeted approach.  And I will also learn from other's PR.

PR #480, opened 2024-01-09
Answer:7 NgRx Notification

jdegand:
This challenge took a lot of research and trial and error to complete.  Check [my repo](https://github.com/jdegand/ngrx-notification) for an alternative solution without ngrx effects.

I used an Angular Material snackbar to display the notification.  

This [article](https://netbasal.com/the-hidden-power-of-injectiontoken-factory-functions-in-angular-d42d5575859b) was a great help.


tomalaforge:
Thanks a lot for taking the time to resolve this challenge. So now we have one example.

I lost my solution and i don{t have the courage to do it again 😅

PR #484, opened 2024-01-09
docs: notification readme rework

jdegand:
You could punch this up even more but I think this is much better explanation of the challenge. 

This would close my issue as well.  

Would still need a PR to touch up the comments inside the challenge code.


tomalaforge (approved):

tomalaforge:
Thanks for the rewrite. I didn't take the time to do it because nobody took the time to answer it since it's a challenge that takes a lot of time and effort.

Thanks a lot for all your help to this repository

PR #486, opened 2024-01-11
fix: update interfaces

jdegand:
Check the api response.  A lot of properties aren't used but this matches the response better.


tomalaforge:
If you want 😅 But I just declared the properties I used. That's just a type and it will be deleted at transpilation.

tomalaforge (approved):

PR #487, opened 2024-01-12
Answer 41: Without Control value accessor

Jesc6417:
Hi,

Good challenge and maybe an answer to my question: Do we really need control value accessor with template driven form?

As I don't use Reactive forms, I've modified the branch to use Template Driven form :)

Here's my solution. I'm not a huge fan about "control value accessor" that's why I don't use it. 
For me, there's a lot of boilerplate that in most of the case, we don't need.

What do you think about this solution?

Just review [feat: solution without control-value-accessor (template-driven-form)](https://github.com/tomalaforge/angular-challenges/pull/487/commits/bf408a28bf24affd7403ccfa7e2c4ce5519159a9)
I don't know why there're two others commits... ^^

Thanks!

### DIFF
--- a/apps/angular/crud/README.md
+++ b/apps/angular/crud/README.md
@@ -8,6 +8,10 @@
 npx nx serve angular-crud
 ```
 
+```bash
+npx nx test angular-crud
+```
+
 ### Documentation and Instruction
 
 Challenge documentation is [here](https://angular-challenges.vercel.app/challenges/angular/5-crud/).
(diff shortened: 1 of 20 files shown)


tomalaforge:
Nice, great initiative. To be honest, I have never used template driven form but i have always wanted to try and use them but I have always been inside companies that don't use them.

But I know that template driven form have a lot less boilerplate. I'm waiting for signal forms 😅

Because i review your code, can you clean your PR and leave only what is relevant to the challenge. There is a lot of code that might belongs to other tests you did. 🙏

Jesc6417:
Ok, I've created a new branch and cherry pick the last commit ;)
I prefer Template driven form for this reason and also for the single responsability. I don't like my "controler" create and manipulate my form. But it's a matter of taste ^^
If you want, we hire people in my company, you could use Template Driven Form ;)

PR #488, opened 2024-01-12
Answer:1

NX-edoardo-colombo:
Esercizio 1


tomalaforge:
Is it your solution or just the beginning? Because you didn't do anything on the card component. 
The goal is to create a highly customizable component.

PR #490, opened 2024-01-12
Translate Challenge 13 and add extra section

kabrunko-dev:
- Translated the challenge 13 to PT
- Added an extra section with a [solution](https://github.com/tomalaforge/angular-challenges/pull/341/files#diff-082bc0154adae46609f52c9e366d4768ae9278dcb495a9a7ec7cb0baf393fb1c) that you liked, @tomalaforge.[^1]
- Fixed a typo in `PageComponent`

[^1]: I thought it would be great to link it so that people could see that's possible to still use some `@Input()` :smile:


kabrunko-dev:
> Thanks for the initiative but I'm not a big fan of doing it like this. I will add a new label like `good solution` or something similar so people can see good solution when they browse into the solution. what do you think?

That would be nice. I'll remove the "Note" section from both files then.

kabrunko-dev:
Done :smile_cat:

jdegand:
> Thanks for the initiative but I'm not a big fan of doing it like this. I will add a new label like `good solution` or something similar so people can see good solution when they browse into the solution. what do you think?

I don't think you should single any solution out.  It is up to the reader to decide.  And if you single out a solution, other solutions will not be looked at as frequently.

kabrunko-dev:
> > Thanks for the initiative but I'm not a big fan of doing it like this. I will add a new label like `good solution` or something similar so people can see good solution when they browse into the solution. what do you think?
> 
> I don't think you should single any solution out. It is up to the reader to decide. And if you single out a solution, other solutions will not be looked at as frequently.

A possible way is use a alternative label as "alternative code" or some text that signals a PR with a different and valid way of doing the challenge.

jdegand:
Then you are going to clutter up the PRs with a bunch of tags.

kabrunko-dev:
> Then you are going to clutter up the PRs with a bunch of tags.

True, true. That's a real possibility.

Well, I think that now it's up to @tomalaforge :thinking:

tomalaforge:
All the responsabilities are on my shoulder now. 

I thought of doing it a while back but I didn't. But having a label to highlight good PR is a good thing. When you want to look up a solution, you don't want to look at all solutions. Most of them are very close but some are better than other or a good and viable alternative. 

For some challenges, you can have multiple ways of doing it, for instance challenges that uses third-party-lib and I try to add a tag to show what is the solution using if you want to focus on a particular way. 

The goal of this repo is not to look at all solution but to see what you can do, how to do it and what is a good solution. 

For bad solution, for instance I add the tag `comments` or `invalid`. 

However I don't know what kind of tag I should use. 🧐

jdegand:
I think most people use just look at the latest solutions.  You could add an `outdated` tag.

kabrunko-dev:
> I still think this is unnecessary. It gives people a shortcut and they won't benefit as much if they don't struggle.

All answers are a shortcut in a way, no? I don't wanna people jump in to an answer quickly as well. However, their effort must reach a point that they don't get frustrated.

There are cases with a ton of answers/PR for a single challenge. Sometimes you can open, i don't know, 3 PRs, but it's the same code line by line. For this reason, sometimes can be hard to find some different implementation.

I like the way PRs are "highlighted" on the `type-challenges` repo where the community chooses which PRs are good giving it a thumbs up (you can see it [here](https://github.com/type-challenges/type-challenges/issues?q=label%3A4+label%3Aanswer+sort%3Areactions-%2B1-desc)).

jdegand:
I think `projection` is really the only challenge flooded with solutions.  I think a voting system is a much better idea than a tag.

tomalaforge:
I really like the upvoting solution. I will implement this and add it somewhere inside the guide. 

And for the hint, I started adding hidden hints and after that I added what the challenge is about inside the information part. It's hard to know what to add. But if you have any advises, don't hesite to fill an issue or implement it.

tomalaforge (approved):

tomalaforge:
@kabrunko-dev @jdegand Thanks for the discussion. 👍 It helps improve the project.

kabrunko-dev:
> @kabrunko-dev @jdegand Thanks for the discussion. 👍 It helps improve the project.

Thank you!

PR #491, opened 2024-01-12
docs: ngrx notification comments

jdegand:
These comments help improve the challenge.


tomalaforge:
I am not a big fan of inline comments for in that case, it can help the user resolving the challenge.

tomalaforge (approved):

PR #493, opened 2024-01-13
Answer:41 solution without control-value-accessor (template-driven-form)

Jesc6417:
Hi Thomas, 

here's my solution without using control-value-accessor.

I've written an article if you wants to share it ;)
https://medium.com/p/2d089fb392c9


tomalaforge:
Thanks a lot for sharing your solution. 🙏
I really need to learn template driven form, but most companies don't use those. And thanks for the article. It can help people to understand better your solution

PR #496, opened 2024-01-14
Answer:1 angular-projection

italloramon:
## My First Challenge 🎉

Hello, this is my first solution. I didn't check other solutions before this because I wanted to try it first without any bias in my mind. However, after that, I will review other solutions to improve mine.

By the way, thanks for this amazing repo ❤️

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,54 @@
 import { Component, OnInit } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import {
+  FakeHttpService,
+  randomCity,
+} from '../../data-access/fake-http.service';
+import { City } from '../../model/city.model';
+import { CardComponent } from '../../ui/card/card.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `
+    <app-card
+      [list]="cities"
+      nameKey="name"
+      customClass="bg-light-blue"
+      (addItem)="onAddItem()"
+      (deleteItem)="onDeleteItem($event)">
+      <img image-header src="assets/img/city.png" width="200px" />
+    </app-card>
+  `,
+  styles: [
+    `
+      ::ng-deep .bg-light-blue {
+        background-color: rgba(0, 0, 250, 0.1);
+      }
+    `,
+  ],
   standalone: true,
-  imports: [],
+  imports: [CardComponent],
 })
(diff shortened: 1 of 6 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -12,21 +7,15 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
+          <app-list-item
+            [name]="item[nameKey]"
+            [id]="item.id"
+            (deleteItem)="onDeleteItem($event)"></app-list-item>
+        }
```
tomalaforge, line 18: this works but it's not very customizable. What will you do if i need to add an icon but only for city Card? and add another button for teacher. you will start to add multiple conditions.
Moreover you will have to modify this component at each new change

Look at `ngTemplateOutlet`

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -18,18 +15,10 @@ import { CardType } from '../../model/card.model';
 export class ListItemComponent {
   @Input() id!: number;
   @Input() name!: string;
-  @Input() type!: CardType;
 
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore,
-  ) {}
+  @Output() deleteItem = new EventEmitter<number>();
```
tomalaforge, line 19: You don't need to pass the ID. `deleteItem` can just trigger an event.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,54 @@
...
-  ngOnInit(): void {}
+  constructor(
+    private http: FakeHttpService,
+    private store: CityStore,
+  ) {}
+
+  ngOnInit(): void {
+    this.http.fetchCities$.subscribe((c) => this.store.addAll(c));
+
+    this.store.cities$.subscribe((c) => (this.cities = c));
```
tomalaforge, line 43: think declarative over imperative. 

prefer to always declare your properties on top of the class. And they will react to changes.

And try to use signals to get use to the new API

tomalaforge:
Great work for your first try. and this is very good to not look at other solutions before. I think you will learn better by trying first, failing and retrying. Copying never brings you somewhere. 

If you have any questions, don't hesitate to ask them here or DM me on X

PR #506, opened 2024-01-19
Answer:43 signal input solution

GRDtechlab:
## Answer:43 Solution for input signal

I do not know what was broken before. But now I cloned repo again in new folder. Now I am submitting my solution for answer:43. Thank you for your time and explanation and everything. I am really going to learn more about git too. Thanks.

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -2,7 +2,7 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
+  input,
   OnChanges,
 } from '@angular/core';
 
@@ -27,15 +27,21 @@ const ageToCategory = (age: number): Category => {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+  // @Input({ required: true }) name!: string;
+  // @Input() lastName?: string;
+  // @Input() age?: string;
+
+  name = input.required<string>();
+  lastName = input<string>('')
+  age = input<string,number>(0, {
+    transform: (v:any) => Number(v)
+  })
 
   fullName = '';
   category: Category = 'Junior';
 
   ngOnChanges(): void {
-    this.fullName = `${this.name} ${this.lastName ?? ''}`;
-    this.category = ageToCategory(Number(this.age) ?? 0);
+    this.fullName = `${this.name()} ${this.lastName()}`;
+    this.category = ageToCategory(Number(this.age()));
   }
 }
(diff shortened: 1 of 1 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -27,15 +27,21 @@ const ageToCategory = (age: number): Category => {
...
 export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+  // @Input({ required: true }) name!: string;
+  // @Input() lastName?: string;
+  // @Input() age?: string;
+
+  name = input.required<string>();
+  lastName = input<string>('')
```
tomalaforge, line 35: use generic only if the value you give is null or undefined, otherwise the type is infered by TS

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -27,15 +27,21 @@ const ageToCategory = (age: number): Category => {
...
+  age = input<string,number>(0, {
+    transform: (v:any) => Number(v)
+  })
```
tomalaforge, line 38: same here

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -27,15 +27,21 @@ const ageToCategory = (age: number): Category => {
...
 
   fullName = '';
   category: Category = 'Junior';
 
   ngOnChanges(): void {
-    this.fullName = `${this.name} ${this.lastName ?? ''}`;
-    this.category = ageToCategory(Number(this.age) ?? 0);
+    this.fullName = `${this.name()} ${this.lastName()}`;
+    this.category = ageToCategory(Number(this.age()));
```
tomalaforge, line 45: use computed, ngOnChange will certainly disappear in upcoming years

tomalaforge:
Nice, you finally did it. But you should not reclone the project but update it regularly. 

I added some comments. Use computed to derive signal state

apps/angular/signal-input/src/app/user.component.ts
GRDtechlab: So, is this correct approach now if I changed based on your suggestions..

`name = input.required<string>('');
  lastName = input<string>()
  age = input<string>()

  fullName = computed(()=> `${this.name()} ${this.lastName()}` );
  category =  computed(() => ageToCategory(Number(this.age())))`

GRDtechlab:
I have updated the code as per my best knowledge as per your suggested comments. I pushed my code into this PR request with commit message number `7662b8c` . Thanks.

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,20 @@ const ageToCategory = (age: number): Category => {
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>('');
+  age = input<number, unknown>(0, {
```
tomalaforge, line 33: you don't need to put the generic type, the TS type checker will infer them for you. It's always better to let the type checker do it for you. Only specify the type when needed

tomalaforge:
That's better, great work

PR #512, opened 2024-01-21
Answer:1 projection solution

SephShom:

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,53 @@
-import { Component, OnInit } from '@angular/core';
+import { AsyncPipe } from '@angular/common';
+import { Component, OnInit, inject, signal } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import {
+  FakeHttpService,
+  randomCity,
+} from '../../data-access/fake-http.service';
+import {
+  CardComponent,
+  CardImageDirective,
+  CardListItemDirective,
+} from '../../ui/card/card.component';
+import { ListItemComponent } from '../../ui/list-item/list-item.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  templateUrl: 'city-card.component.html',
+  styles: [
+    `
+      :host {
+        background-color: rgba(0, 0, 250, 0.1);
+        width: max-content;
+        height: max-content;
+      }
+    `,
+  ],
   standalone: true,
-  imports: [],
+  imports: [
+    CardComponent,
+    CardImageDirective,
+    CardListItemDirective,
+    ListItemComponent,
+    AsyncPipe,
+  ],
 })
 export class CityCardComponent implements OnInit {
-  constructor() {}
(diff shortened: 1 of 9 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,55 @@
...
+    <div class="flex w-fit flex-col gap-3 rounded-md border-2 border-black p-4">
+      <ng-content select="app-card-image"></ng-content>
 
       <section>
-        <app-list-item
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track $index) {
```
tomalaforge, line 34: be careful with $index. If the list changed, the $index is not efficient.
you should always try to track a property of your list and use $index for list that doesn't change

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,55 @@
...
+          <ng-template
+            [ngTemplateOutlet]="rowTemplate.templateRef"
+            [ngTemplateOutletContext]="{ $implicit: item }"></ng-template>
+        }
```
tomalaforge, line 38: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,55 @@
...
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+export class CardComponent<T> {
+  @Input() list: T[] | null = null;
+  @Output() actionClicked = new EventEmitter();
+  @ContentChild(CardListItemDirective) rowTemplate!: CardListItemDirective<T>;
```
tomalaforge, line 54: 👍

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,53 @@
...
+    AsyncPipe,
+  ],
 })
 export class CityCardComponent implements OnInit {
-  constructor() {}
+  private http = inject(FakeHttpService);
+  private store = inject(CityStore);
 
-  ngOnInit(): void {}
+  public cities$ = signal(this.store.cities$);
```
tomalaforge, line 40: Don't do that. don't put an observable inside a signal. 
use `toSignal()` to transform an obs to signal

tomalaforge:
great work but be careful, don't put any observable inside a signal

apps/angular/projection/src/app/ui/card/card.component.ts
SephShom: I really appreciate your observation. At this case, generic T has no explicit properties, it is correct if track with whole item?:  @for (item of list; track item)

tomalaforge: You can do <T extends {id: number)> so each element has to have an id.

PR #515, opened 2024-01-21
Answer:5 crud ngrx solution

SephShom:

### DIFF
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,11 +1,5 @@
 import { CommonModule } from '@angular/common';
-import {
-  Component,
-  OnInit,
-  WritableSignal,
-  inject,
-  signal,
-} from '@angular/core';
+import { Component, OnInit, WritableSignal, inject } from '@angular/core';
 import { MatProgressBarModule } from '@angular/material/progress-bar';
 import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
 import { MatSnackBarModule } from '@angular/material/snack-bar';
@@ -52,15 +46,17 @@ import { TaskComponent } from './ui/task/task.component';
   styles: [],
 })
 export class AppComponent implements OnInit {
-  private stodos = signal<Task[]>([]);
-  public get todos(): Task[] {
-    return this.stodos();
-  }
   public loaderService = inject(LoaderService);
   private store = inject(Store);
 
+  private stodos = this.store.selectSignal(selectList) as WritableSignal<
+    Task[]
+  >;
+  public get todos(): Task[] {
+    return this.stodos();
+  }
+
   ngOnInit(): void {
     this.store.dispatch(taskActions.loadTasksAction());
-    this.stodos = this.store.selectSignal(selectList) as WritableSignal<Task[]>;
   }
 }
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,51 +1,73 @@
 import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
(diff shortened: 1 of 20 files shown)


apps/angular/crud/src/app/store/tasks/task.effects.ts
```diff
@@ -0,0 +1,89 @@
...
+          ),
+        );
+      }),
+    ),
+  );
+
+  delTaskEffect = createEffect(() =>
+    this.actions$.pipe(
+      ofType(taskActions.delTaskAction),
+      switchMap((action) => {
```
tomalaforge, line 66: be careful with switchMap on delete, post, put. 
You will only cancel the request, not the change in the database.

apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,73 @@
...
-          userId: todo.userId,
-        }),
-        {
-          headers: {
-            'Content-type': 'application/json; charset=UTF-8',
-          },
+  ngOnInit(): void {
+    this.store.dispatch(taskActions.loadTasksAction());
+    this.sysSubs.add(
+      this.store.select(getTaskListSelector).subscribe({
```
tomalaforge, line 65: don't code imperatively. 

you have selectSignal in Ngrx, so you can immedialy do stodo = selectSIgnal(mySelector). 

Avoid setting class properties inside ngOnInit.

SephShom: Great!

apps/angular/crud/src/app/ui/loader/loader.service.ts
```diff
@@ -0,0 +1,9 @@
+import { Injectable, signal } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class LoaderService {
+  //public isLoading$: BehaviorSubject<boolean> = new BehaviorSubject(false);
+  public isLoading$ = signal<boolean>(false);
```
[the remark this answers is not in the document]
SephShom: No more $, it would be good to have another char to signals, maybe

apps/angular/crud/src/app/store/tasks/task.effects.ts
SephShom: Strongly important, thank you!

apps/angular/crud/src/app/app.component.ts
```diff
@@ -46,28 +51,16 @@ import { TaskComponent } from './ui/task/task.component';
...
     this.store.dispatch(taskActions.loadTasksAction());
-    this.sysSubs.add(
-      this.store.select(getTaskListSelector).subscribe({
-        next: (res) => this.stodos.set(res),
-        error: () => {
-          this.stodos.set([]);
-        },
-      }),
-    );
+    this.stodos = this.store.selectSignal(selectList) as WritableSignal<Task[]>;
```
tomalaforge, line 64: you don't have to initialize it inside ngoninit. You can do it immediatly  l.55

This is why signal are reactive, they will reactive to any changes

apps/angular/crud/src/app/store/tasks/task.effects.ts
```diff
@@ -63,9 +63,9 @@ export class TasksEffects {
   delTaskEffect = createEffect(() =>
     this.actions$.pipe(
       ofType(taskActions.delTaskAction),
-      switchMap((action) => {
+      exhaustMap((action) => {
         return this.service.delEntity(action.task).pipe(
-          switchMap((data) => {
+          mergeMap((data) => {
```
tomalaforge, line 68: why do you do a mergeMap with a of?

simply do map

SephShom: I do mergeMap because i want to fatten to different actions as a response. Update list and show a confirmation message. Maybe this is not the best approach, I'll be glad to know another option

PR #518, opened 2024-01-21
Answer:4 Typed ContextOutlet 

stillst:


tomalaforge:
Nice solution. I love this challenge. The DX is so great

stillst:
> Nice solution. I love this challenge. The DX is so great

Thanks for the challenge, i started to use the solution at work for typing templates in ng-select. It helps a lot.

PR #521, opened 2024-01-22
Answer:8 pure pipe

LKL2017:
Done and I think if we can add more content to this challenge since it is called **challenge**, even though it is labeled as easy. For example, maybe some tests?


apps/angular/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,17 @@
...
 
 @Component({
   standalone: true,
-  imports: [NgFor],
+  imports: [NgFor, HeavyComputationPipe],
   selector: 'app-root',
   template: `
     <div *ngFor="let person of persons; let index = index">
-      {{ heavyComputation(person, index) }}
+      {{ person | heavyCompute: index }}
```
tomalaforge, line 11: 👍

PR #523, opened 2024-01-22
Answer:1 projection solution

ahmed-aouinti:

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,64 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, inject } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
+import { switchMap, tap } from 'rxjs';
+import { CityStore } from '../../data-access/city.store';
+import {
+  FakeHttpService,
+  randomCity,
+} from '../../data-access/fake-http.service';
+import { City } from '../../model/city.model';
+import { CardListItemDirective } from '../../ui/card/card-list-item.directive';
+import { CardComponent } from '../../ui/card/card.component';
+import { ListItemComponent } from '../../ui/list-item/list-item.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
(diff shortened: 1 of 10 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,46 @@
...
-        *ngIf="type === CardType.TEACHER"
-        src="assets/img/teacher.png"
-        width="200px" />
-      <img
-        *ngIf="type === CardType.STUDENT"
-        src="assets/img/student.webp"
-        width="200px" />
+    <ng-content select="[image]"></ng-content>
+    <section>
+      @for (item of list; track item) {
```
tomalaforge, line 18: `track: item` is a dangerous default track option. You should use it only if the list is not updated (like adding or deleting element) , prefer using a specific property like id

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,46 @@
...
-  CardType = CardType;
-
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore,
-  ) {}
+export class CardComponent<T> {
+  @Input() list: T[] | null = null;
+  @Output() addedItem = new EventEmitter<void>();
+  @ContentChild(CardListItemDirective, { read: TemplateRef })
```
tomalaforge, line 40: 👍

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,64 @@
...
+    this.http.fetchCities$.pipe(
+      tap((cities) => {
+        this.store.addAll(cities);
+      }),
+      switchMap(() => {
+        return this.store.cities$;
+      }),
+    ),
+    { initialValue: [] as City[] },
+  );
```
tomalaforge, line 55: I don't really like this kind of implementation. 

Loading the cities just be separate from the state cities. 

You should call `fetchCities` inside the constructor or ngoninit and get the city from the store. More maintainable, easier to read.

tomalaforge:
Great work

PR #524, opened 2024-01-22
Answer:43 inputs to signal inputs

SephShom:

Answer:43

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -2,8 +2,9 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
-  OnChanges,
+  computed,
+  input,
+  numberAttribute,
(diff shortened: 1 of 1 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,20 @@ const ageToCategory = (age: number): Category => {
...
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required();
+  lastName = input();
```
tomalaforge, line 32: I think you need to give a type otherwise you have lastName is of type `signal<unknown, unknown>`

SephShom: Thank you, I could check it. Fixed on next commit

tomalaforge:
Perfect

PR #526, opened 2024-01-22
Answer:40 web workers

stillst:


apps/performance/christmas-web-worker/src/app/app.component.ts
```diff
@@ -10,23 +17,62 @@ import { UnknownPersonComponent } from './unknown-person/unknown-person.componen
...
   `,
   host: {
     class: `flex flex-col h-screen w-screen bg-[#1f75c0]`,
   },
 })
 export class AppComponent {
-  private heavyCalculationService = inject(HeavyCalculationService);
+  private readonly heavyCalculationService = inject(HeavyCalculationService);
+
+  private readonly webWorkerLoadingPercentage: WritableSignal<number> =
```
tomalaforge, line 38: you don't have to type your property, TS will infer it for you. easier to read and TS is always right. 

You can type it if you want to narrow it.

tomalaforge:
Nice work, What is easy ? I struggle implementing the service worker when I did the challenge because it's not very intuitive on how the SW is working. 

Great work 👍

apps/performance/christmas-web-worker/src/app/app.component.ts
stillst: we have a convention at work to type all properties, so it it probably a habit :) 
fixed

stillst:
> Nice work, What is easy ? I struggle implementing the service worker when I did the challenge because it's not very intuitive on how the SW is working.
> 
> Great work 👍

Thanks! Well, there's the schematic and basic [example](https://angular.io/guide/web-worker) from the challenge instructions. I think that it is enough for simple logic. But if i come across a non-standard complex task, i will most probably need to spend some time on research :-)

apps/performance/christmas-web-worker/src/app/app.component.ts
tomalaforge: Yes many companies have this habits, but I don't really get it. It add more code, and TS is more accurate than we are. 😅

tomalaforge:
> > Nice work, What is easy ? I struggle implementing the service worker when I did the challenge because it's not very intuitive on how the SW is working.
> > Great work 👍
> 
> Thanks! Well, there's the schematic and basic [example](https://angular.io/guide/web-worker) from the challenge instructions. I think that it is enough for simple logic. But if i come across a non-standard complex task, i will most probably need to spend some time on research :-)

So I might need to create a more complex challenge. But the issue is, most people don't do complex challenges 😅

stillst:
> > > Nice work, What is easy ? I struggle implementing the service worker when I did the challenge because it's not very intuitive on how the SW is working.
> > > Great work 👍
> > 
> > 
> > Thanks! Well, there's the schematic and basic [example](https://angular.io/guide/web-worker) from the challenge instructions. I think that it is enough for simple logic. But if i come across a non-standard complex task, i will most probably need to spend some time on research :-)
> 
> So I might need to create a more complex challenge. But the issue is, most people don't do complex challenges 😅

IMHO current difficulty of the challenge is perfect and somebody want to try web worker this challenge is great start. :-)

apps/performance/christmas-web-worker/src/app/app.component.ts
stillst: Well as far as i remember the main points were:
1 when you make the review, you can see types in the code, so you have more info about the code and better understanding of it.
2 when you write code you should explicitly specify types to make sure that functions return exactly what you expect them to return

PR #531, opened 2024-01-24
Answer:1 Fixing Code

dpinalopez:


apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="border-2 border-black rounded-md p-4 w-fit flex flex-col gap-3" [class]="customClass">
+    <ng-content select="[cardImage]"></ng-content>
+
+    <section>
+        <ng-container>
+            @for (item of list; track item; ) {
```
tomalaforge, line 6: you should prefer to target a property of item. item is a back default.

apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -0,0 +1,13 @@
...
+            <ng-template [ngTemplateOutlet]="itemRef" [ngTemplateOutletContext]="{$implicit: item}"></ng-template>
+            }
```
tomalaforge, line 8: 👍

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,15 @@
...
-  ) {}
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Input() itemRef!: TemplateRef<{ $implicit: T }>;
```
tomalaforge, line 14: you can use ContentChild to project a template inside your child

apps/angular/projection/src/app/data-access/teacher.store.ts
```diff
@@ -1,23 +1,22 @@
...
 import { Teacher } from '../model/teacher.model';
 
 @Injectable({
   providedIn: 'root',
 })
 export class TeacherStore {
-  private teachers = new BehaviorSubject<Teacher[]>([]);
-  teachers$ = this.teachers.asObservable();
+  private _teachers = signal<Teacher[]>([]);
+  teachers = computed(this._teachers);
```
tomalaforge, line 9: you can use _teachers.asReadonly() to create a readonly signal

tomalaforge:
That's good work. 👍

apps/angular/projection/src/app/ui/card/card.component.html
dpinalopez: I don't get it, Any example?

tomalaforge: item is quite useless. because since we use immutable data, the list will be rerender each time we delete or add an element. 
You should prefer $index as a default if we do not move/replace/sort element in the list, which we do. 
BUt the best is to target a property like item.id for instance

dpinalopez: Thanks!

apps/angular/projection/src/app/ui/card/card.component.ts
dpinalopez: I dit it in my local code Thanks!

apps/angular/projection/src/app/data-access/teacher.store.ts
dpinalopez: I dit it in my local code Thanks!

PR #542, opened 2024-01-27
Answer: 2 effect vs reducer vs selector

amosISA:


apps/ngrx/effect-selector/src/app/app.component.ts
```diff
@@ -1,72 +1,29 @@
...
-        border-color: black;
-        padding: 2px;
-      }
-    `,
-  ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class AppComponent implements OnInit {
-  private store = inject(Store);
+  readonly #store: Store<AppStore> = inject(Store<AppStore>);
```
tomalaforge, line 20: you don't have to type your store

apps/ngrx/effect-selector/src/app/app.component.ts
```diff
@@ -1,72 +1,29 @@
...
 
-  activities$ = this.store.select(selectActivities);
+  activities = this.#store.selectSignal(
+    ActivitySelectors.selectActivitiesWithTeachers,
+  );
```
tomalaforge, line 24: 👍

apps/ngrx/effect-selector/src/app/store/activity/activity.selectors.ts
```diff
@@ -8,3 +11,32 @@ export const selectActivities = createSelector(
...
+    };
+
+    return activities.map(mapActivityWithTeachers);
+  },
+);
+
+export const ActivitySelectors = {
+  selectActivities,
+  selectActivitiesWithTeachers,
+};
```
tomalaforge, line 42: 👍

apps/ngrx/effect-selector/src/app/app.component.ts
amosISA: why not?

tomalaforge: it's always better to let typescript infer the type.

PR #546, opened 2024-01-28
Answer: 1 Signal based solution

gesiud:


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,44 @@
...
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Output() add = new EventEmitter<void>();
+  @ContentChild('rowRef', { read: TemplateRef })
```
tomalaforge, line 42: prefer using a directive to avoid mabic string

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,18 @@
...
-  ) {}
-
-  delete(id: number) {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.deleteOne(id);
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.deleteOne(id);
-    }
-  }
+  id = input<number>();
```
tomalaforge, line 16: you don't need to pass an id, you can use the output as an event trigger

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -21,20 +37,31 @@ import { CardComponent } from '../../ui/card/card.component';
...
-    private http: FakeHttpService,
-    private store: TeacherStore,
-  ) {}
+    effect(
+      () => {
+        this.store.addAll(fetchTeachers());
+      },
+      { allowSignalWrites: true },
+    );
+  }
```
tomalaforge, line 58: I'm not a big fan of this implementation. Using `allowSignalWrites` is often a code smell.

You can still use subscribe when needed. Using only signal is not the way to go. You should use signal when they fit, otherwise stay with rxjs. The goal is not to over-engineer your code

PR #547, opened 2024-01-29
Answer 43: Input Signals Solution

BillyOkoth:
refactoring form using @Input decorator to input signals

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -4,6 +4,7 @@ import {
   Component,
   computed,
   input,
+  numberAttribute,
 } from '@angular/core';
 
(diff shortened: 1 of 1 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,17 @@ const ageToCategory = (age: number): Category => {
...
-  fullName = '';
-  category: Category = 'Junior';
-
-  ngOnChanges(): void {
-    this.fullName = `${this.name} ${this.lastName ?? ''}`;
-    this.category = ageToCategory(Number(this.age) ?? 0);
-  }
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
```
tomalaforge, line 32: you should give a default value when it's not required. To avoid using `?? '' ` later in the code

PR #556, opened 2024-02-02
fix: english json

jdegand:
The english json had some french included.  This should fix it.


tomalaforge:
indeed, but already fixed.

PR #562, opened 2024-02-03
Answer 1: content projection

NonieBenks:

### DIFF
--- a/apps/angular/projection/src/app/ui/card/card.component.ts
+++ b/apps/angular/projection/src/app/ui/card/card.component.ts
@@ -1,8 +1,4 @@
-import { NgFor, NgIf } from '@angular/common';
-import { Component, Input } from '@angular/core';
-import { randStudent, randTeacher } from '../../data-access/fake-http.service';
-import { StudentStore } from '../../data-access/student.store';
-import { TeacherStore } from '../../data-access/teacher.store';
+import { Component, EventEmitter, Input, Output } from '@angular/core';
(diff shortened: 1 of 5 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -12,50 +8,46 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
-        width="200px" />
-
+      <ng-content select="card-image"></ng-content>
       <section>
-        <app-list-item
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
```
tomalaforge, line 13: 👍

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -1,35 +1,17 @@
...
 
 @Component({
   selector: 'app-list-item',
   template: `
     <div class="border-grey-300 flex justify-between border px-2 py-1">
       {{ name }}
-      <button (click)="delete(id)">
-        <img class="h-5" src="assets/svg/trash.svg" />
-      </button>
+      <ng-content select="list-button"></ng-content>
```
tomalaforge, line 9: The button can stay on this component. That's a generic button

However you should use a ng-content instead of name. It's not very customizable

PR #565, opened 2024-02-06
docs(pt): translate challenge 44

kabrunko-dev:
And remove "New" badge from challenge 43 which is about `Signal Input`.


docs/src/content/docs/pt/challenges/angular/43-signal-input.md
```diff
@@ -6,7 +6,6 @@ challengeNumber: 43
 command: angular-signal-input
 sidebar:
   order: 16
-  badge: New
```
tomalaforge, line 9: I need to add this inside the schematic 👍

PR #569, opened 2024-02-08
Answer:44 view transition challenge

wandri:
Learned new transition view method

Some feedback:

- I didn't figure what was the most efficient way to play with `view-transition-name`.

   - I tried `[style.viewTransitionName]="'logo'+postId()"` but I would prefer to keep the inputs for business logic.

   - Finally I used css only (and classes). But not a huge fan of it.

- I lost some time on the management of the display to finish to use `scrollIntoView({behavior: 'instant'})`. This is why I have decreased the size of the cover in order to play with it easly.


tomalaforge:
Thanks a lot for your solution.

I prefer your solution over mine. You used on transition API to apply the `view-transition-name` to a different card, which is better than using the router Event. I have a small lag when I start my transition and you don't, 

Thanks for taking the time to do it. 🙏

PR #571, opened 2024-02-08
Answer:5

simonzea:

### DIFF
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,51 +1,58 @@
 import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
-import { Component, OnInit } from '@angular/core';
+import {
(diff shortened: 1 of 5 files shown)


apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,58 @@
...
-          },
-        },
-      )
-      .subscribe((todoUpdated: any) => {
-        this.todos[todoUpdated.id - 1] = todoUpdated;
-      });
+  update(todo: User) {
+    const todoClone = { ...todo };
+    todoClone.title = randText();
+    this.appService.updateUser(todoClone).subscribe((todoUpdated: User) => {
```
tomalaforge, line 46: You need to be careful with subscribing. I always prefer to add an unsubscribe method even if I know that http request complete. You never know what the network will be.

tomalaforge:
Good, nice work

PR #575, opened 2024-02-09
Answer:44 View transition

jdegand:
My solution is not best practice. Imperative solution. Duplicated selectors and animations. Absolute pixel values. Not mobile responsive.   This [PR](https://github.com/tomalaforge/angular-challenges/pull/569) is a superior solution.   

However, I learned a lot about the inner workings of the view transition api and I think there is some value in this solution, as it shows how hard it can be to do this manually for a lot of items.   The various levels of the readme is slightly deceptive as strategies for a singular animation doesn't necessarily translate across many items.

### DIFF
--- a/apps/angular/view-transition/src/styles.scss
+++ b/apps/angular/view-transition/src/styles.scss
@@ -12,55 +12,16 @@
 }
 */
 
+@media (prefers-reduced-motion) {
+
+    ::view-transition-group(*),
+    ::view-transition-old(*),
+    ::view-transition-new(*) {
+        animation: none !important;
+    }
+}
+
 /*
-.main-header {
-    view-transition-name: main-header;
-    contain: layout;
-}
-
-.main-header-text {
-    view-transition-name: main-header-text;
-    contain: layout;
-    width: fit-content;
-}
-
-@media (min-width: 530px) {
-    .main-nav {
-        view-transition-name: main-nav;
-        contain: layout;
-    }
-}
-
-@keyframes fade-in {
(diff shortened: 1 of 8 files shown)


apps/angular/view-transition/src/styles.scss
```diff
@@ -2,3 +2,187 @@
...
+    animation: 1500ms cubic-bezier(0.4, 0, 0.2, 1) slide-to-right-blue;
+}
+
+.blue::view-transition-new(angular3) {
+    display: none;
+}
+
+.blue::view-transition-new(profil3) {
+    display: none;
+}
```
tomalaforge, line 174: That's what we want to avoid. Because on this example, we only have 3 thumbails, but in a real project we can have a lot more 😅

and also you can write them like:
```css
::view-transition-new(angular1a) {
    animation: 1500ms cubic-bezier(0.4, 0, 0.2, 1) slide-to-left-2;
}

::view-transition-new(profil2a),
::view-transition-new(angular2a) {
    animation: 1500ms cubic-bezier(0.4, 0, 0.2, 1) slide-to-left-2;
}
```

since the css property is the same

tomalaforge:
Thanks for your solution. The fact that the `view-transition-name` must be unique make the challenge really hard. And you need to master the animation as well to have a nice effect.
But I love seeing the solution of others. I had a really hard time on this one also, and this [PR](https://github.com/tomalaforge/angular-challenges/pull/569) is indeed the best solution so far.

PR #579, opened 2024-02-12
 Answer:5 Update with the solution using signal

abhishekw1:
Added Signals in the service file and accessed in the component file. 
Tried to keep the business logic at the service file.

All Senior please comment on this, so that I can improve next time.


apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,52 @@
...
-    <div *ngFor="let todo of todos">
+    <!-- 
+    // this is optional  
+    @if (isError) {
+      <p class="error">{{ isError }}</p>
+    } -->
+    @if (isLoading) {
+      <mat-spinner></mat-spinner>
+    }
+    <div *ngFor="let todo of todos$">
```
tomalaforge, line 19: I don't think this is working. ngFor is not imported. You should use the new control flow

apps/angular/crud/src/app/services/todo.service.ts
```diff
@@ -0,0 +1,83 @@
...
+
+@Injectable({ providedIn: 'root' })
+export class TodoService {
+  _url = 'https://jsonplaceholder.typicode.com/todos/';
+
+  appSignal = signal<appModel>({
+    todos: [],
+    // isError: null, //optional
+    isLoading: false,
+  });
```
tomalaforge, line 14: I think I prefer two separates signals.

apps/angular/crud/src/app/services/todo.service.ts
```diff
@@ -0,0 +1,83 @@
+import { HttpClient } from '@angular/common/http';
+import { Injectable, signal } from '@angular/core';
+import { appModel, todo } from '../todo.model';
+import { randText } from '@ngneat/falso';
+
+@Injectable({ providedIn: 'root' })
+export class TodoService {
```
tomalaforge, line 7: The service is good.

tomalaforge:
This is ok. You just need to use your signals outside of the service without using an effect.

apps/angular/crud/src/app/app.component.ts
abhishekw1: It was working, but now updated with @for

apps/angular/crud/src/app/services/todo.service.ts
abhishekw1: Yes added new two separated Computed signals

apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,52 @@
...
+  isLoading!: boolean;
 
+  constructor(private todoService: TodoService) {
+    effect(() => {
+      // const { todos, isError, isLoading } = this.todoService.appSignal();
+      const { todos, isLoading } = this.todoService.appSignal();
+      this.todos$ = todos;
+      // this.isError = isError;
+      this.isLoading = isLoading;
+    });
```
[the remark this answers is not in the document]
abhishekw1: I have added effect so that the value changes of signals so it will update the values of todos$. but now got it, so we can use computed signals and declared Signal at starting assigned values.


Thank you for your valuable review.

apps/angular/crud/src/app/services/todo.service.ts
abhishekw1: Thank you

abhishekw1:
added updated solution as per the feedback

apps/angular/crud/src/app/app.component.ts
```diff
@@ -13,31 +13,25 @@ import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
...
+    }
   `,
   styles: [],
 })
 export class AppComponent implements OnInit {
-  todos$!: todo[];
-  // isError!: string | null;
-  isLoading!: boolean;
+  todos$: Signal<todo[]> = this.todoService.todoSignal;
+  isLoading: Signal<boolean> = this.todoService.loadingSignal;
```
tomalaforge, line 31: 👍

PR #580, opened 2024-02-12
Answer:31 Module to standalone

bill-griffin:
* Used the Angular schematic `nx g @angular/core:standalone` from https://angular.io/guide/standalone-migration
* The first step of the schematic, `convert-to-standalone` worked perfectly.
* Ran the second step, `prune-ng-modules`, and nothing happened (reason: see below)
* Manually converted all `RouterModule.forChild(...)` into `Route[]` constants
* Manually moved `providers: [provideToken(...)]` to the Route[] constants
* Used `default` exports to avoid `.then(...)` in `loadChildren` and `loadComponent`
* Noticed that I had to delete all `RouterModule.forChild(...)` and `providers: [...]` from the NgModules before the `prune-ng-modules` would be deleted by the schematic
* I had one Module that wouldn't get deleted, so I manually deleted (sorry, forgot which one)
* The final step, `standalone-bootstrap`, worked fine, just needed a tweak for a missing import.


tomalaforge:
Great work and amazing PR description. 
I added a 👍 to rank it to the first place so people can get inspiration of your description for their own projects.

PR #581, opened 2024-02-15
feat(solution): projection

ipopa144:
Answer:1

### DIFF
--- a/apps/angular/projection/src/app/component/student-card/student-card.component.ts
+++ b/apps/angular/projection/src/app/component/student-card/student-card.component.ts
@@ -1,40 +1,62 @@
+import { AsyncPipe } from '@angular/common';
 import { Component, OnInit } from '@angular/core';
-import { FakeHttpService } from '../../data-access/fake-http.service';
+import { Observable } from 'rxjs';
+import {
+  FakeHttpService,
+  randStudent,
+} from '../../data-access/fake-http.service';
 import { StudentStore } from '../../data-access/student.store';
-import { CardType } from '../../model/card.model';
 import { Student } from '../../model/student.model';
 import { CardComponent } from '../../ui/card/card.component';
+import { ListItemComponent } from '../../ui/list-item/list-item.component';
 
 @Component({
   selector: 'app-student-card',
   template: `
     <app-card
-      [list]="students"
-      [type]="cardType"
-      customClass="bg-light-green"></app-card>
+      [list]="students$ | async"
+      (add)="addStudent()"
+      class="bg-light-green">
+      <img src="assets/img/student.webp" width="200px" />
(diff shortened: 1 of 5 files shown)


wandri:
Hi @ipopa144, the title of your PR should start with `Answer:1` for example `Answer:1 projection` instead of `feat(solution): projection`

apps/angular/projection/src/app/component/student-card/student-card.component.ts
```diff
@@ -1,40 +1,62 @@
...
     <app-card
-      [list]="students"
-      [type]="cardType"
-      customClass="bg-light-green"></app-card>
+      [list]="students$ | async"
+      (add)="addStudent()"
+      class="bg-light-green">
+      <img src="assets/img/student.webp" width="200px" />
+
+      <ng-template #rowRef let-student>
```
svenson95, line 22: The magic string `#rowRef` could be replaced by a directive, here is a nice stackoverflow answer with reasons why we should avoid them. [What is wrong with magic strings?](https://softwareengineering.stackexchange.com/questions/365339/what-is-wrong-with-magic-strings)

One good point:

> You may coincidentally use the same magic string in two places, when really they are different things, so if you did a Find & Replace, and changed both, one of them could break while the other worked.

>

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,48 @@
...
-        width="200px" />
+      <ng-content select="img"></ng-content>
 
       <section>
-        <app-list-item
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        <ng-container *ngFor="let item of list">
```
tomalaforge, line 20: use the new control flow 💪

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,48 @@
...
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @ContentChild('rowRef', { read: TemplateRef }) rowTemplate!: TemplateRef<{
+    $implicit: T;
+  }>;
```
tomalaforge, line 47: prefect directive over magic string.

tomalaforge:
Thanks @wandri and @svenson95 for the review. I'm looking forward to have more review from the community. 

Good job @ipopa144, I added some comments to code in modern angular

PR #584, opened 2024-02-15
Answer:45 - React in Angular | Author solution

wandri:
Possible solution, step by step:

1. Install the dependencies
2. Add `"jsx": "react"` in `compilerOptions` in `tsconfig.json` in order to allow the compiler to read correctly the react extensions.
3. To render the react component, switch `post.component.ts` to `post.component.tsx`
4. Use the React API `createRoot` and `render` to display the react component.
5. For the design with Tailwind, add the `tsx` extension in `tailwind.condig.js`

To go further:

- Destroy the component correctly with for example
```
  ngOnDestroy() {
    ReactDOM.unmountComponentAtNode(this.containerRef.nativeElement);
  }
```


apps/angular/react-in-angular/src/app/react/post.component.tsx
```diff
@@ -0,0 +1,48 @@
...
+export class PostComponent implements AfterViewInit, OnChanges {
+  @ViewChild('react', { static: true }) containerRef!: ElementRef;
+
+  post = input<Post | undefined>(undefined);
+  isSelected = input<boolean>(false);
+  @Output() selectPost = new EventEmitter<void>();
+  root?: Root;
+
+  ngOnChanges(): void {
+    this.render();
```
tomalaforge, line 25: Why did you add it here as well?

jdegand: @tomalaforge Are you referring to `root` ?

tomalaforge: @jdegand no, why do we need to rerender at each change?

jdegand: Well, we might not need it in this case since the posts are not changing.  I think it is typically done to ensure that the Angular component will re-render if there is any change in the React component.  React and Angular have different change detection strategies.   React avoids unnecessary changes (virtual DOM) so there could be a mismatch.

tomalaforge: Thanks for the explanation. I have never used React.

PR #588, opened 2024-02-15
Answer:8 easy pipe solution

ahmed-aouinti:


apps/angular/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,17 @@
...
+import { HeavyComputationPipe } from './shared/HeavyComputation.pipe';
 
 @Component({
   standalone: true,
-  imports: [NgFor],
   selector: 'app-root',
   template: `
     <div *ngFor="let person of persons; let index = index">
-      {{ heavyComputation(person, index) }}
+      {{ person | heavy: index }}
```
tomalaforge, line 10: very good, you could have used the new control flow syntax, but this is not the goal of this challenge.

PR #590, opened 2024-02-16
Answer 5: Crud NGRX Solution 

abhishekw1:
Added Crud Solution Using the NGRX with the latest syntax without signal

### DIFF
--- a/apps/angular/crud/src/app/app.component.ts
+++ b/apps/angular/crud/src/app/app.component.ts
@@ -1,51 +1,60 @@
 import { CommonModule } from '@angular/common';
-import { HttpClient } from '@angular/common/http';
 import { Component, OnInit } from '@angular/core';
-import { randText } from '@ngneat/falso';
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { Store } from '@ngrx/store';
+import { Observable } from 'rxjs';
+import { TODO } from './app.service';
+import {
+  deleteTodoAction,
+  getAllTodosAction,
+  updateTodoAction,
+} from './store/app.actions';
+import { isError, isLoading, selectAllTodos } from './store/app.selectors';
+
+// import { GetAllTodoActions } from './store/app.actions';
 
 @Component({
   standalone: true,
-  imports: [CommonModule],
+  imports: [CommonModule, MatProgressSpinnerModule],
   selector: 'app-root',
   template: `
-    <div *ngFor="let todo of todos">
+    @if (isError$ | async) {
+      <p class="error">{{ isError$ | async }}</p>
+    }
+    @if (isLoading$ | async) {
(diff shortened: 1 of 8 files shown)


apps/angular/crud/src/app/store/app.effects.ts
```diff
@@ -0,0 +1,64 @@
...
+        ),
+      ),
+    ),
+  );
+
+  deleteTodo = createEffect(() =>
+    this.actions$.pipe(
+      ofType(deleteTodoAction),
+      switchMap(({ id }) => {
+        const did = id;
```
tomalaforge, line 51: why do you create an intermediate variable?

apps/angular/crud/src/app/store/app.actions.ts
```diff
@@ -0,0 +1,53 @@
...
+);
+export const successdeleteTodoAction = createAction(
+  SUCCESSDELETEDTODO,
+  props<{ id: number; loading: boolean }>(),
+);
+
+export const errorAction = createAction(
+  ERRORACTION,
+  props<{ error: string }>(),
+);
```
tomalaforge, line 44: you can use createGroupAction.

apps/angular/crud/src/app/store/app.actions.ts
```diff
@@ -0,0 +1,53 @@
...
+// import { Action } from '@ngrx/store';
+import { TODO } from '../app.service';
+
+export const GETALLTODO = '[Todo] GetAllTodo';
+export const UPDATETODO = '[Todo] UpdateTodo';
+export const SUCCESSUPDATEDTODO = '[Todo] SuccessfullyUpdatedTodo';
+export const DELETETODO = '[Todo] DeleteTodo';
+export const SUCCESSDELETEDTODO = '[Todo] SuccessfullyDeletedTodo';
+export const ADDEDALLTODO = '[Todo] AddedAllTodo';
+export const ERRORACTION = '[Todo] Error Occured';
```
tomalaforge, line 11: That's very old school to create constant for each action. This was needed in the old ngrx store.

apps/angular/crud/src/app/app.component.ts
```diff
@@ -1,51 +1,60 @@
...
 
   ngOnInit(): void {
-    this.http
-      .get<any[]>('https://jsonplaceholder.typicode.com/todos')
-      .subscribe((todos) => {
-        this.todos = todos;
-      });
+    this.todos$ = this.store.select(selectAllTodos);
+    this.isLoading$ = this.store.select(isLoading);
+    this.isError$ = this.store.select(isError);
```
tomalaforge, line 45: you can instanciate it directly inside the field initializer

And you can use selectSignal to have a signal

tomalaforge:
Nice work, but you could modernize the code. use signal and the new ngrx store API

PR #592, opened 2024-02-16
Answer:8 intermediate pipe solution

ahmed-aouinti:


tomalaforge:
The goal of the challenge is to create a very generic pipe. You just need to plug your function into the pipe to get the memoize advantage of the pipe.

PR #593, opened 2024-02-16
Answer:8 hard pipe solution

ahmed-aouinti:


tomalaforge:
This challenge is more of a TS challenge. The goal is to create a generic pipe that works for any utility libraries. Not for only one function. And the pipe must be strongly typed.

PR #594, opened 2024-02-16
Answer 8: added easy pipe solution

abhishekw1:
Some Pointers that I I have learned form Solution Article provided by @tomalaforge:
- Pipes are a way to transform data before it is displayed in the template.
- They are memoized, meaning that if the inputs to the pipe have not changed, the last computed value will be returned.
- They can be pure or impure. Pure pipes only re-execute if their inputs change, while impure pipes re-execute on every change detection cycle.
- They can be used with multiple arguments.
- They are created using the @Pipe decorator.
- They can be injected into any component, directive, or service.


tomalaforge:
Nice summary 💪

But I think pipe will kind of disappear with signal.

PR #597, opened 2024-02-17
Answer:45 React

jdegand:
I am comfortable in both, but using them together still felt weird.  

This [repo](https://github.com/aligneddev/angular-plus-react) really helped me out.   

I used lifecycle methods in the wrapper component.  Computed or effect may also work.  

Initially, I tried using a React CDN.  I had some difficulty, but it should be possible.

In my research, I have seen injection tokens being used for access to libraries.  Maybe that is also a possibility.     

Check out [this article](https://johnfewell.com/blog/angularjs-angular17/) for an interesting Angularjs and Angular 17 combo.


tomalaforge:
Yes I agree, doesn't feel very natural. And it's not using a react standalone component, because the component needs to be inside an angular component. 
I don't know if you can use this technique for a real enterprise project.

jdegand:
[This Stack Overflow question](https://stackoverflow.com/questions/29849114/can-angular-and-react-play-together) mentions that React components in Angular started as a way to gain performance.   Angular has made some big progress in performance and I feel like React has stagnated a little.  React, by itself, is not really recommended -> Next and Remix are.

tomalaforge:
Next yes but I don't know about remix. In our current entreprise proejct, we are moving away from Remix.

PR #599, opened 2024-02-18
Answer 21: Solution For Anchor Navigation for both home and foo route

abhishekw1:


apps/angular/anchor-scrolling/src/app/nav-button.component.ts
```diff
@@ -1,10 +1,12 @@
 /* eslint-disable @angular-eslint/component-selector */
 import { Component, Input } from '@angular/core';
+import { RouterLinkWithHref } from '@angular/router';
 @Component({
   selector: 'nav-button',
   standalone: true,
+  imports: [RouterLinkWithHref],
   template: `
-    <a [href]="href">
+    <a [routerLink]="href" [fragment]="section">
```
tomalaforge, line 9: 👍

PR #601, opened 2024-02-18
Answer 43: solution for input decorator to signal input

abhishekw1:


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -2,8 +2,8 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
-  OnChanges,
+  computed,
+  input,
 } from '@angular/core';
 
 type Category = 'Youth' | 'Junior' | 'Open' | 'Senior';
```
svenson95, line 9: Just a little thing, but i would move that `ageToCategory` method inside the Category class as a static method, since it returns a Category object. That could be a little bit cleaner, i always try to avoid stuff outside the component class, expect for static values.

tomalaforge: Even for big projects, I don't use that OOP pattern. I prefer having functions than classes. 
I hate the new keyword. I prefer having to call a pure function.

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +19,31 @@ const ageToCategory = (age: number): Category => {
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
+  age = input<string>();
```
tomalaforge, line 32: you can use the transform method to transform it to number

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +19,31 @@ const ageToCategory = (age: number): Category => {
...
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
```
tomalaforge, line 31: when the input is not required, use a default value.

PR #603, opened 2024-02-18
Answer:43 signal input solution

svenson95:

### DIFF
--- a/apps/angular/signal-input/src/app/models/category.model.ts
+++ b/apps/angular/signal-input/src/app/models/category.model.ts
@@ -1,15 +1,10 @@
-export enum CategoryType {
-  Youth = 'Youth',
-  Junior = 'Junior',
-  Open = 'Open',
-  Senior = 'Senior',
-}
+export type CategoryType = 'YOUTH' | 'JUNIOR' | 'OPEN' | 'SENIOR';
 
 export class Category {
-  static ageToCategory(age: number): CategoryType {
-    if (age < 10) return CategoryType.Youth;
-    else if (age < 18) return CategoryType.Junior;
-    else if (age < 35) return CategoryType.Open;
-    return CategoryType.Senior;
+  static fromAge(age: number): CategoryType {
+    if (age < 10) return 'YOUTH';
+    else if (age < 18) return 'JUNIOR';
+    else if (age < 35) return 'OPEN';
+    return 'SENIOR';
   }
 }
--- a/apps/angular/signal-input/src/app/models/category.model.ts
+++ b/apps/angular/signal-input/src/app/models/category.model.ts
@@ -1,0 +1,15 @@
+export enum CategoryType {
+  Youth = 'Youth',
+  Junior = 'Junior',
+  Open = 'Open',
+  Senior = 'Senior',
+}
+
+export class Category {
+  static ageToCategory(age: number): CategoryType {
+    if (age < 10) return CategoryType.Youth;
+    else if (age < 18) return CategoryType.Junior;
+    else if (age < 35) return CategoryType.Open;
+    return CategoryType.Senior;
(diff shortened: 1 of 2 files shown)


apps/angular/signal-input/src/app/models/category.model.ts
```diff
@@ -0,0 +1,15 @@
...
+}
+
+export class Category {
+  static ageToCategory(age: number): CategoryType {
+    if (age < 10) return CategoryType.Youth;
+    else if (age < 18) return CategoryType.Junior;
+    else if (age < 35) return CategoryType.Open;
+    return CategoryType.Senior;
+  }
+}
```
tomalaforge, line 15: From your code, we can see that you come from a backend OOP language like Java. Am I right? 😅

In TS, I always avoid enums. Enums were created for java developers but don't fit well in TS. Union string or map type is better. 

I don't see the value of the static function inside a class. A pure function inside a util file feels more natural and more functional. Less boilerplate

But both implementations work. Thanks for showing your implementation.

svenson95: No but i kind of like the Java syntax 😄 

I'll give it a try and refactor this, thanks for your advice!

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -2,40 +2,30 @@ import { TitleCasePipe } from '@angular/common';
...
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
-
-  fullName = '';
-  category: Category = 'Junior';
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
+  age = input<string>();
```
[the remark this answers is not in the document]
svenson95: nice hint thank you

apps/angular/signal-input/src/app/models/category.model.ts
svenson95: @tomalaforge Maybe thats an idea for a new typescript challenge like _Avoid Enums_?

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -2,40 +2,32 @@ import { TitleCasePipe } from '@angular/common';
...
-  fullName = '';
-  category: Category = 'Junior';
-
-  ngOnChanges(): void {
-    this.fullName = `${this.name} ${this.lastName ?? ''}`;
-    this.category = ageToCategory(Number(this.age) ?? 0);
-  }
+  fullName = computed(() => `${this.name()} ${this.lastName()}`);
+  category = computed(() =>
+    this.age() ? Category.fromAge(this.age()!) : CATEGORY_DEFAULT,
```
svenson95, line 31: @tomalaforge I changed the function name here, only to show why i like to use that static method in a class approach. It is more boilerplate but it reads like a Sentence. Get the `Category.fromAge(this.age()!)`.

apps/angular/signal-input/src/app/models/category.model.ts
tomalaforge: Yes we can create a challenge, that's right.
you can look at this video from TS guru: https://www.youtube.com/watch?v=jjMbPt_H3RQ&t=480s&ab_channel=MattPocock

apps/angular/signal-input/src/app/user.component.ts
tomalaforge: If you prefer it and you feel more comfortable with it. that's good. the goal is to have a scalable and maintainable application so that you can easily refactor it and understand it in a couple of months.
I know that i prefer functional programming, but that's my personal preference. In the end, there is no gain in performance.

PR #606, opened 2024-02-19
docs(pt): translate remaining pipe challenges

kabrunko-dev:


docs/src/content/docs/pt/challenges/angular/9-pipe-wrapFn.md
```diff
@@ -0,0 +1,35 @@
...
+challengeNumber: 9
+command: angular-pipe-intermediate
+blogLink: https://medium.com/ngconf/boost-your-apps-performance-by-wrapping-your-functions-inside-a-pipe-7e889a901d1d
+sidebar:
+  order: 103
+---
+
+## Informação
+
+Este é o primeiro de três desafios `@Pipe()`, sendo o objetivo dominar **pipes** em Angular.
```
svenson95, line 14: I'm not sure, but this should not be "primeiro" but "segundo de três desafios"? if so, then in 10-pipe-utility it should be "**terceiro** de três desafios" as well. Or something like that 😄

kabrunko-dev: You're right @svenson95. Thank you!

tomalaforge (approved):

PR #608, opened 2024-02-19
Answer:34

satutama:
- Extracted the input component out.
- Extracted the row component out.
- Added changeDetectionStrategy.OnPush for PersonListComponent, RandomComponent
- Use @For block
- Use of KeyboardEvent instead of type Any and event.key === "Enter", because keycode is deprecated


apps/performance/default-onpush/src/app/name-input.component.ts
```diff
@@ -0,0 +1,43 @@
...
+    </mat-form-field>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class NameInputComponent {
+  @Output() newNameEvent = new EventEmitter<string>();
+
+  label = '';
+
+  handleKey(event: KeyboardEvent) {
```
svenson95, line 37: You can use the `(keydown.enter)` event in your template, so you don't have to check if the pressed key is `Enter`. More information about this [here](https://www.digitalocean.com/community/tutorials/angular-binding-keyup-keydown-events).

satutama: Refactored, and thanks for that! Something new I learnt today 😄

apps/performance/default-onpush/src/app/name-input.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    CDFlashingDirective,
+  ],
+  template: `
+    <mat-form-field class="w-full" cd-flash>
+      <input
+        placeholder="Add one member to the list"
+        matInput
+        type="text"
+        [(ngModel)]="label"
+        (keydown.enter)="handleEnter()" />
```
tomalaforge, line 27: 👍

tomalaforge:
Great work

PR #611, opened 2024-02-19
Answer:43

dpinalopez:

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -2,8 +2,8 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
   OnChanges,
+  input,
 } from '@angular/core';
 
 type Category = 'Youth' | 'Junior' | 'Open' | 'Senior';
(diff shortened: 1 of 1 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -27,15 +27,14 @@ const ageToCategory = (age: number): Category => {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
-
+  name = input.required<string>();
+  lastName = input<string>();
+  age = input(0, { transform: (value: string) => Number(value) });
```
tomalaforge, line 32: there are built-in transform function: you can use numberAttribute

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -27,15 +27,14 @@ const ageToCategory = (age: number): Category => {
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
-
+  name = input.required<string>();
+  lastName = input<string>();
```
tomalaforge, line 31: you can use a default value when the input is not required.

PR #613, opened 2024-02-20
Answer:45

BillyOkoth:
1. use of a directive to intergrate react component in angular


tomalaforge:
I like your solution. I prefer it over the other one. Using a directive to wrap the react component is a nice move. Great work and thanks for implementing it.

BillyOkoth:
Thank you very much for the feedback and for creating a platform for challenges

PR #615, opened 2024-02-20
style: red button unintended font-color black

svenson95:
I guess two buttons have the wrong font color, because there are similar buttons with a different font color:

- landing page of Angular Challenges(https://angular-challenges.vercel.app/) the subscribe button
Bildschirmfoto 2024-02-20 um 01 31 59

- all other pages (https://angular-challenges.vercel.app/guides/getting-started/) the sponsor button
Bildschirmfoto 2024-02-20 um 01 35 44

### DIFF
--- a/docs/src/components/ActionButtonFooter.astro
+++ b/docs/src/components/ActionButtonFooter.astro
@@ -51,8 +51,6 @@ const { data } = await getEntry('i18n', lang);
   .button-sponsor {
     background-color: red;
     border-color: red;
-    color: white;
+    color: white !important;
   }
-
-
 </style>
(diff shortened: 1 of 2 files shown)


docs/src/components/ActionButtonFooter.astro
```diff
@@ -51,8 +51,6 @@ const { data } = await getEntry('i18n', lang);
   .button-sponsor {
     background-color: red;
     border-color: red;
-    color: white;
+    color: white !important;
```
tomalaforge, line 54: indeed this is working fine for that time, but it's often better to use CSS variables because some colors differ from dark and light themes.

tomalaforge (approved):

PR #621, opened 2024-02-24
Answer:43 signal input 

stillst:


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +19,19 @@ const ageToCategory = (age: number): Category => {
...
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  $name = input.required<string>({ alias: 'name' });
+  $lastName = input('', { alias: 'lastName' });
+  $age = input(0, { alias: 'age', transform: Number });
```
tomalaforge, line 33: For transform, i didn't know that Number was working. 👍

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +19,19 @@ const ageToCategory = (age: number): Category => {
...
 
-  fullName = '';
-  category: Category = 'Junior';
-
-  ngOnChanges(): void {
-    this.fullName = `${this.name} ${this.lastName ?? ''}`;
-    this.category = ageToCategory(Number(this.age) ?? 0);
-  }
+  $fullName = computed(() => `${this.$name()} ${this.$lastName()}`);
+  $category = computed(() => ageToCategory(Number(this.$age())));
```
tomalaforge, line 36: why do you use Number again here, `age` should already be of type number

apps/angular/signal-input/src/app/user.component.ts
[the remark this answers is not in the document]
stillst: I was just about to ask for your opinion on this syntax 😅
We've started using it at work to spot errors in templates like 
```
*ngIf="signalVariable"
```
At first glance, everything seems correct, but actually we need this
```
*ngIf="signalVariable()"
```

apps/angular/signal-input/src/app/user.component.ts
stillst: you right, missed it

PR #623, opened 2024-02-25
Answer:6 Permissions solution

svenson95:

### DIFF
--- a/apps/angular/permissions/src/app/has-permission.guard.ts
+++ b/apps/angular/permissions/src/app/has-permission.guard.ts
@@ -1,0 +1,44 @@
+import { inject, Injectable } from '@angular/core';
+import { CanMatchFn, Route, Router, UrlTree } from '@angular/router';
+
+import { Role } from './user.model';
+import { UserStore } from './user.store';
+
+@Injectable({ providedIn: 'root' })
+export class HasPermissionService {
+  private readonly router = inject(Router);
+  private readonly userStore = inject(UserStore);
+
+  canMatch(route: Route): boolean | UrlTree {
+    const isAdmin: boolean = route.data?.['isAdmin'] ?? false;
+    const roles: Role[] = route.data?.['roles'] ?? [];
+    return this.hasPermission$(isAdmin, roles);
+  }
+
+  private hasPermission$(
+    isAdmin: boolean,
+    accessRolesList: Role[],
+  ): boolean | UrlTree {
+    const isLoggedIn = this.userStore.isUserLoggedIn();
+
+    if (!isLoggedIn) {
+      return this.router.parseUrl('no-user');
+    }
+
+    if (accessRolesList.length > 0) {
+      return this.userStore.hasAnyRole(accessRolesList);
+    }
+
+    if (isAdmin) {
+      return this.userStore.isAdmin();
+    }
+
+    return false;
+  }
+}
+
+export const HasPermissionGuard: CanMatchFn = (
+  next: Route,
+): boolean | UrlTree => {
+  return inject(HasPermissionService).canMatch(next);
+};
--- a/apps/angular/permissions/src/app/has-permission.guard.ts
+++ b/apps/angular/permissions/src/app/has-permission.guard.ts
@@ -1,44 +1,24 @@
-import { inject, Injectable } from '@angular/core';
-import { CanMatchFn, Route, Router, UrlTree } from '@angular/router';
+import { inject } from '@angular/core';
+import { Router, UrlTree } from '@angular/router';
 
+import { Observable, of } from 'rxjs';
 import { Role } from './user.model';
 import { UserStore } from './user.store';
 
-@Injectable({ providedIn: 'root' })
-export class HasPermissionService {
-  private readonly router = inject(Router);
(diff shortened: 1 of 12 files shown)


apps/angular/permissions/src/app/has-permission.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Role } from './user.model';
+import { UserStore } from './user.store';
+
+@Injectable({ providedIn: 'root' })
+export class HasPermissionService {
+  private readonly router = inject(Router);
+  private readonly userStore = inject(UserStore);
+
+  canMatch(route: Route): boolean | UrlTree {
+    const isAdmin: boolean = route.data?.['isAdmin'] ?? false;
```
tomalaforge, line 13: The goal of using a function is to avoid this hack. use function parameters instead of using the data property of the route.

apps/angular/permissions/src/app/has-role.directive.ts
```diff
@@ -0,0 +1,50 @@
...
+  selector: '[appHasRole], [appHasRoleIsAdmin]',
+  providers: [provideDestroyService()],
+})
+export class HasRoleDirective {
+  private readonly templateRef = inject(TemplateRef<unknown>);
+  private readonly viewContainer = inject(ViewContainerRef);
+  private readonly store = inject(UserStore);
+
+  @Input('appHasRole') role: Role[] = [];
+  @Input('appHasRoleIsAdmin') isAdmin = false;
```
tomalaforge, line 25: you can use signal input if you want to play with signals

apps/angular/permissions/src/app/information.component.ts
```diff
@@ -1,24 +1,25 @@
...
     <h2 class="mt-10 text-xl">Information Panel</h2>
     <!-- admin can see everything -->
-    <div>visible only for super admin</div>
-    <div>visible if manager</div>
-    <div>visible if manager and/or reader</div>
-    <div>visible if manager and/or writer</div>
-    <div>visible if client</div>
+    <div *appHasRoleIsAdmin="true">visible only for super admin</div>
+    <div *appHasRole="['MANAGER']">visible if manager</div>
+    <div *appHasRole="['MANAGER', 'READER']">
```
tomalaforge, line 14: 👍

apps/angular/permissions/src/app/login.component.ts
```diff
@@ -35,29 +39,20 @@ import { UserStore } from './user.store';
...
+  readonly admin = admin;
+  readonly manager = manager;
+  readonly reader = reader;
+  readonly writer = writer;
+  readonly readerAndWriter = readerAndWriter;
+  readonly client = client;
+  readonly everyone = everyone;
+
+  select(role: User): void {
+    this.userStore.select(role);
```
tomalaforge, line 56: a nicer way to do something similar would be to be:
select accept a union type, than you create a map between your union type and the wanted object. 👍

apps/angular/permissions/src/app/routes.ts
```diff
@@ -1,14 +1,51 @@
...
   {
     path: 'enter',
-    loadComponent: () =>
-      import('./dashboard/admin.component').then(
-        (m) => m.AdminDashboardComponent,
-      ),
+    canMatch: [() => inject(HasPermissionGuard).canMatch()],
+    data: {
+      isAdmin: true,
+    },
```
tomalaforge, line 20: this should be a function param, data is not type safe

apps/angular/permissions/src/app/has-permission.guard.ts
svenson95: yes you're right, updated

apps/angular/permissions/src/app/has-role.directive.ts
svenson95: done

apps/angular/permissions/src/app/login.component.ts
svenson95: updated in [here](https://github.com/tomalaforge/angular-challenges/pull/623/commits/8c40ac42f938be2c057283e2bffec844e01bda88)

apps/angular/permissions/src/app/routes.ts
svenson95: done

svenson95:
@tomalaforge thanks for the feedback!

apps/angular/permissions/src/app/has-role.directive.ts
```diff
@@ -0,0 +1,50 @@
...
+        const isAdmin = this.store.isAdmin();
+        isAdmin ? this.addTemplate() : this.clearTemplate();
+      } else if (this.role) {
+        this.store.hasAnyRole(this.role)
+          ? this.addTemplate()
+          : this.clearTemplate();
+      } else {
+        this.addTemplate();
+      }
+    });
```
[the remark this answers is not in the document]
tomalaforge: yes but if `IsAdmin` is false, then the effect doesn't listen to the signal.
I need to dig more on that subject. I haven't play a lot with effect yet.

PR #631, opened 2024-02-27
Answer:46 Simple animations

jdegand:
I made animations slower than the videos.

### DIFF
--- a/apps/angular/simple-animations/src/app/app.component.ts
+++ b/apps/angular/simple-animations/src/app/app.component.ts
@@ -1,3 +1,11 @@
+import {
+  animate,
+  query,
+  stagger,
+  style,
+  transition,
+  trigger,
+} from '@angular/animations';
 import { Component } from '@angular/core';
 
 @Component({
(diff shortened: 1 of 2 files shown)


apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -17,9 +25,32 @@ import { Component } from '@angular/core';
       }
     }
   `,
+  animations: [
+    trigger('slideIn', [
```
svenson95, line 29: Personally i prefer to create a component animations file `app.animation.ts` and keep the animation outside of the component, to keep the component as simple as possible (less lines of code). But that's totally fine too.

apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -17,9 +25,32 @@ import { Component } from '@angular/core';
...
+            ),
+          ]),
+        ]),
+      ]),
+    ]),
+  ],
   template: `
     <div class="mx-20 my-40 flex gap-5">
-      <section>
+      <!-- The posts could be inside *ngFor loop but for an animation to work a trackBy function is required  -->
```
svenson95, line 52: That's interesting, nice hint 👍 i'll try it with a `@for` loop to see if it works.

apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -17,9 +25,32 @@ import { Component } from '@angular/core';
       }
     }
   `,
+  animations: [
+    trigger('slideIn', [
+      transition(':enter', [
+        style({ transform: 'translateX(-100%)' }),
+        animate('1000ms ease-in', style({ transform: 'translateX(0%)' })),
```
svenson95, line 32: Maybe it's possible to use `transform: '*'` instead of `transform: 'translateX(0%)'`. 
More information about the `*` token [here](https://angular.io/api/animations/style).

svenson95 (commented):
Nice work! 🚀

apps/angular/simple-animations/src/app/app.component.ts
jdegand: Yes, `auto-styling` works here.  `auto-styling` gives more flexibility. If the `transform` were to change, you wouldn't have to update the `animate`.   I'll leave the code as it is, but I don't know any negatives to using `auto-styling`.

apps/angular/simple-animations/src/app/app.config.ts
```diff
@@ -1,5 +1,8 @@
 import { ApplicationConfig } from '@angular/core';
+import { provideAnimations } from '@angular/platform-browser/animations';
 
 export const appConfig: ApplicationConfig = {
-  providers: [],
+  providers: [provideAnimations()],
 };
+
+// provideAnimationsAsync won't enable animations on load
```
tomalaforge, line 8: That's an interesting information. But indeed, we should use async only to delay the load of the package. If we know that we need the animation right away, that's normal to load the package immedialty

PR #634, opened 2024-02-28
Answer:13 Highly Customizable CSS solution

svenson95:

---

My solution don't use any css variables. In my opinion it's better to keep the text styling within the text component.


apps/angular/styling/src/app/text.component.ts
```diff
@@ -1,16 +1,36 @@
...
+
+    :host-context(.error) p {
+      font-size: 30px;
+      color: red;
+    }
+
+    :host-context(.warning) p {
+      font-size: 25px;
+      color: orange;
+    }
```
tomalaforge, line 33: the goal is to use css variables to be as customizable as possible.

svenson95: done

PR #636, opened 2024-02-28
Answer:32 Change Detection Bug solution

svenson95:

### DIFF
--- a/apps/angular/bug-cd/src/app/main-navigation.component.ts
+++ b/apps/angular/bug-cd/src/app/main-navigation.component.ts
@@ -1,6 +1,7 @@
-import { AsyncPipe, NgFor, NgIf } from '@angular/common';
-import { Component, Input, inject } from '@angular/core';
+import { Component, Input, computed, inject } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { RouterLink, RouterLinkActive } from '@angular/router';
+
 import { FakeServiceService } from './fake.service';
 
 interface MenuItem {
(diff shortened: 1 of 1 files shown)


apps/angular/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -39,26 +40,18 @@ export class NavigationComponent {
...
   `,
   host: {},
 })
 export class MainNavigationComponent {
-  private fakeBackend = inject(FakeServiceService);
-
-  readonly info$ = this.fakeBackend.getInfoFromBackend();
+  private readonly fakeBackend = inject(FakeServiceService);
+  private readonly info = toSignal(this.fakeBackend.getInfoFromBackend());
+  readonly menus = computed(() => this.getMenu(this.info() ?? ''));
```
tomalaforge, line 52: 👍

apps/angular/bug-cd/src/app/main-navigation.component.ts
```diff
@@ -11,16 +12,16 @@ interface MenuItem {
 @Component({
   selector: 'app-nav',
   standalone: true,
-  imports: [RouterLink, RouterLinkActive, NgFor],
+  imports: [RouterLink, RouterLinkActive],
   template: `
-    <ng-container *ngFor="let menu of menus">
+    @for (menu of menus; track menu.path) {
```
tomalaforge, line 17: 👍

tomalaforge:
Did you understand why we had an issue?

svenson95:
> Did you understand why we had an issue?

I think so. The problem was that `routerLinkActive` triggers a change detection cycle under the hood. Without a trackby function in `*ngFor` loop all items have been removed from the DOM and added again. Through this `routerLinkActive` was called again, thus we had a infinite loop of rerenderings.

tomalaforge:
no exactly.
trackby hides the issue because you compare a value inside your list and not an instance id. So as you said, the list is not rerender and we don't recall `routerLinkActive` 
but the issue is before that. We call a function inside our template that recreate a list each time. So we have a new list at each CD cycle, which is a bad design. 
So if you store your list inside a variable, even without the trackby, the solution will be solved.

svenson95:
ah okay, i thought that template function call is just bad practice but not the reason of this problem. 

thank's a lot for the feedback!

PR #638, opened 2024-02-29
feat(challenge 47): typescript enums vs union types

svenson95:
New typescript challenge as discussed in this [PR](https://github.com/tomalaforge/angular-challenges/pull/603#discussion_r1498261040).


docs/src/content/docs/challenges/typescript/47-enums-vs-union-types.md
```diff
@@ -0,0 +1,67 @@
...
+- [Should You Use Enums or Union Types in Typescript?](https://www.bam.tech/article/should-you-use-enums-or-union-types-in-typescript)
+- [Typescript has unions, so are enums redundant?](https://stackoverflow.com/questions/40275832/typescript-has-unions-so-are-enums-redundant)
+- [Tidy TypeScript: Prefer union types over enums](https://fettblog.eu/tidy-typescript-avoid-enums/)
+
+## Statement
+
+The goal of this challenge is to refactor the enums `Difficulty` & `Direction`.
+
+- Refactor `Difficulty` to const enum.
+- Refactor `Direction` to union type.
```
tomalaforge, line 67: I would prefer to transform everything inside union type and map type for runtime. What do you thing?

because like you stated, const enum are a workaround and I'm not a fan at all of this. 
I created a eslint rule to avoid enum. See challenge 27😅

tomalaforge:
Hey you are on fire. 🔥

I have to change the number of my next challenge each time you submit one 😂

But that's a good one. And one I wanted to create as well, So thanks a lot for doing it 🙏

docs/src/content/docs/challenges/typescript/47-enums-vs-union-types.md
svenson95: I prefer the `const enum` way because of the improved maintainability. But yes when you've created a eslint rule against it i'm changing the tasks to union types only. 

My Opinion is that the advantage is bigger of better maintaining against the disadvantages of like using workarounds. I've seen the video about typescript enums but if you look in the comments, there are many people who keep using enums.

tomalaforge: Yes but even the TS doc is saying to avoid them: https://www.typescriptlang.org/docs/handbook/enums.html#const-enums 😅

svenson95:
@tomalaforge since this is a topic which probably is not important in many cases - when performance is really important, should this challenge not be a hard one?

Of course performance is important, but in my experience usability and accessibility is often more important and often neglected too.

docs/src/content/docs/challenges/typescript/47-enums-vs-union-types.md
svenson95: Update challenge statement done 👍 read the docs again I've changed a lot, added your last comment about _ts docs says avoid const enums_ as well.

tomalaforge (approved):

tomalaforge:
very nice 👍

PR #640, opened 2024-03-03
Answer:1 Projection

Coderbegner:

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,44 @@
 import { Component, OnInit } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import {
+  FakeHttpService,
+  randomCity,
+} from '../../data-access/fake-http.service';
+import { CardType } from '../../model/card.model';
+import { City } from '../../model/city.model';
+import { CardComponent } from '../../ui/card/card.component';
+import { ListItemComponent } from '../../ui/list-item/list-item.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  templateUrl: 'city-card.component.html',
   standalone: true,
-  imports: [],
+  styles: [
+    `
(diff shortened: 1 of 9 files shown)


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,44 @@
...
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  templateUrl: 'city-card.component.html',
   standalone: true,
-  imports: [],
+  styles: [
+    `
+      ::ng-deep .bg-light-green {
```
svenson95, line 18: Try to avoid `::ng-deep` if not necessary, as in this challenge. With your implementation the class `bg-light-green` is declared globally, so if you use this class somewhere else in your application this styles will be assigned.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,44 @@
...
 })
 export class CityCardComponent implements OnInit {
-  constructor() {}
+  constructor(
+    private http: FakeHttpService,
+    private store: CityStore,
+  ) {}
+  cities: City[] = [];
+  cardType = CardType.CITY;
+  id!: number;
```
svenson95, line 32: `id!: number;` is not defined? i don't see where it gets assigned.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,32 @@
...
 })
 export class CardComponent {
   @Input() list: any[] | null = null;
-  @Input() type!: CardType;
   @Input() customClass = '';
+  @Output() add = new EventEmitter<void>();
 
-  CardType = CardType;
+  @ContentChild('rowRef', { read: TemplateRef })
+  rowTemplate!: TemplateRef<{ $implicit: any }>;
```
tomalaforge, line 25: you can use a directive instead of magic string

PR #644, opened 2024-03-04
Answer:46

satutama:
- Refactor the list items with @for block
- provided provideAnimations()
- Added slideIn animation and stagger animation

### DIFF
--- a/apps/angular/simple-animations/src/app/app.component.ts
+++ b/apps/angular/simple-animations/src/app/app.component.ts
@@ -21,7 +21,7 @@ import { ANIMATIONS } from './app.animation';
   `,
   template: `
     <div class="mx-20 my-40 flex gap-5">
-      <section [@slideIn]>
+      <section @slideIn>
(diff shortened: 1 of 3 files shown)


apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -19,7 +21,7 @@ import { Component } from '@angular/core';
   `,
   template: `
     <div class="mx-20 my-40 flex gap-5">
-      <section>
+      <section [@slideIn]>
```
svenson95, line 24: Are these brackets needed because you pass an array `ANIMATIONS`? Nice way of adding the animations by the way, i've always imported the animations seperate. This seems a bit cleaner 👍

satutama: hey nice catch, thanks! I was binding it to some value but decided not to. But forgot to remove the binding syntax.
I have fixed that.

PR #646, opened 2024-03-05
Answer 46 simple animations async

svenson95:
I forgot that i wanted to try `provideAnimationsAsync()`, had to change a bit of my animation as well to make it work.

### DIFF
--- a/apps/angular/simple-animations/src/app/app.animations.ts
+++ b/apps/angular/simple-animations/src/app/app.animations.ts
@@ -1,0 +1,35 @@
+import {
+  animate,
+  keyframes,
+  query,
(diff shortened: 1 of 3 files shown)


tomalaforge:
In this case, you should not use the async provider, because you want the animation immediately. The async provider is great if you want to load the animation package later in your application.

svenson95:
Yes this is not the best use case for that. I'll take this as a lesson for my next challenge creations, to check out similar methods or providers in this case, and find a good example as a use case.

svenson95:
That sounds good, i'll add a bonus task in the challenge for this async provider subject.

PR #647, opened 2024-03-05
Style button hover

svenson95:
- Added hover styling for following buttons in landing page
  - Get Started
  - Go to the latest Challenge
  - Give a star
  - Sponsor the Project (in footer)
- Added hover styling for buttons in content pages (e. g. challenge page) in the right sidebar
  - Email subscription
  - Give a star
  - Sponsor
- Added hover styling for buttons in article footer
  - Community solutions*
  - Author's solution
  - Article
  - ...

On the angular challenges landing page is a subscribe button, which i fixed a week ago or so - font color in dark mode wasn't updated to white.

I noticed that there is a button hover styling on the subscribe button but not on other similar buttons. So i added a hover styling for the similar buttons and fixed the hover styling of the subscribe button.

### DIFF
--- a/docs/src/components/SubscriptionForm.astro
+++ b/docs/src/components/SubscriptionForm.astro
@@ -85,12 +85,9 @@ const { data } = await getEntry('i18n', lang ?? 'en');
       text-decoration: none;
       font-size: var(--sl-text-sm);
       background: var(--sl-color-text-accent);
-      transition: var(--button-transition);
 
       &:hover {
         top: unset;
-        transform: var(--button-transform);
-        transition: var(--button-transition);
       }
     }
   }
--- a/docs/src/components/SubscriptionForm.astro
+++ b/docs/src/components/SubscriptionForm.astro
@@ -88,6 +88,7 @@ const { data } = await getEntry('i18n', lang ?? 'en');
 
       &:hover {
         top: unset;
+        opacity: unset;
       }
     }
   }
--- a/docs/src/components/SubscriptionForm.astro
+++ b/docs/src/components/SubscriptionForm.astro
@@ -85,6 +85,13 @@ const { data } = await getEntry('i18n', lang ?? 'en');
       text-decoration: none;
       font-size: var(--sl-text-sm);
       background: var(--sl-color-text-accent);
+      transition: var(--button-transition);
+
+      &:hover {
+        top: unset;
+        transform: var(--button-transform);
+        transition: var(--button-transition);
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


docs/src/components/SubscriptionForm.astro
```diff
@@ -85,6 +85,13 @@ const { data } = await getEntry('i18n', lang ?? 'en');
       text-decoration: none;
       font-size: var(--sl-text-sm);
       background: var(--sl-color-text-accent);
+      transition: var(--button-transition);
```
svenson95, line 88: Added `transition` here to also show the out-animation, without these, only the hover is animated but not on unhovering.

docs/src/components/SubscriptionForm.astro
```diff
@@ -85,6 +85,13 @@ const { data } = await getEntry('i18n', lang ?? 'en');
...
+
+      &:hover {
+        top: unset;
```
svenson95, line 91: `top: unset` because the current implementation is from a library i guess, so i couldn't find the styling in this project. Moving the button up with `top` did work but the transition doesn't work for the `top` attribute, now the button moves up through `transform: translateY(-1px)`.

docs/src/styles/custom-css.css
```diff
@@ -50,6 +53,18 @@
   margin: 0px 2px;
 }
 
+a.action, 
+.article-footer > a, 
+.action-footer > a.action-button {
```
svenson95, line 58: - `a.action` are all landing-page start buttons, get started, go to latest challenge ...
- `.article-footer > a` explains itself 😄 
- the `.action-footer` is in the right sidebar on a content page

docs/src/styles/custom-css.css
```diff
@@ -73,6 +88,10 @@
   align-items: center;
   margin-top: 0;
   background-color: var(--sl-color-accent-high);
+
+  &.action {
```
svenson95, line 92: In light mode this button has black font-color while hover on it.

Bildschirmfoto 2024-03-05 um 18 15 32

tomalaforge:
Thanks a lot for improving the UX of the page. 
But I don't know how I feel about the button moving up on hover 😅. I think I prefer a simple hover. too much animations for me 😅

svenson95:
Yeah that's fine, i've just add these animations because the email subscribe button has it already. So that all buttons have roughly the same styling. 

Do you want me to change the subscribe button hover styling, remove that moving up and just `opacity: 0.9`? The subscribe button have this opacity 0.9 currently, that alone looks fine for me too.

tomalaforge:
Indeed, I haven't seen that the button was moving. I think I still prefer without the small animation. Too many animations for me 👍
Thanks for doing that

svenson95:
Requested changes done

tomalaforge (approved):

svenson95:
@tomalaforge then I misunderstood something, i'll add the hover effect again in a new PR.

PR #648, opened 2024-03-05
Answer: 47 - enums vs union

satutama:
- refactored Difficulty and Direction as union type 
- type Direction has lower case union type so it makes sense to use a mapped type
- added mapped type DirectionLabelMap
- refactored the difficultyLabel since we just need to show the signal value
- refactored the directionLabel with conditional operator, I think it is cleaner this way.

### DIFF
--- a/apps/typescript/enums-vs-union-types/src/app/app.component.ts
+++ b/apps/typescript/enums-vs-union-types/src/app/app.component.ts
@@ -10,6 +10,14 @@ const directionLabels: DirectionLabelMap = {
   right: 'Right',
 };
 
+/*  line 6 to 11 can be refactored to use Record type like below.
+    I commented this out since the challenge is to use a Mapped type */
+
+// const directionLabels: Record<Direction, string> = {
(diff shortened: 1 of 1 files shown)


apps/typescript/enums-vs-union-types/src/app/app.component.ts
```diff
@@ -1,14 +1,14 @@
...
+type Difficulty = 'Easy' | 'Normal';
+type Direction = 'left' | 'right';
 
-enum Direction {
-  LEFT = 'left',
-  RIGHT = 'right',
-}
+type DirectionLabelMap = { [key in Direction]: string };
+
+const directionLabels: DirectionLabelMap = {
```
tomalaforge, line 8: you can also use a record here

satutama: yes indeed! But then I will not use a mapped type which is one of the goal of the challenge. 
I added the record solution but commented it out for those who are wondering.
Thank you!

PR #652, opened 2024-03-06
Answer:47 Enum union

jdegand:
Check my previous commits for the commented out implemenation I had.

I didn't implement here, but a POJO with `as const` would be another great replacement for an enum.   

I am not as comfortable using mapped types as the other methods.   Mapped Types require extra steps and the implementation here kind of eliminates some of that.

### DIFF
--- a/apps/typescript/enums-vs-union-types/src/app/app.component.ts
+++ b/apps/typescript/enums-vs-union-types/src/app/app.component.ts
@@ -94,9 +94,10 @@ export class AppComponent {
     }
   });
 
-  // either string or key seems to work
+  // either string or key works
+  // if not left / right / undefined -> error
   readonly directionLabel = computed<string>(() => {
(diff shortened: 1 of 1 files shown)


apps/typescript/enums-vs-union-types/src/app/app.component.ts
```diff
@@ -54,28 +81,28 @@ enum Direction {
...
   readonly difficultyLabel = computed<string>(() => {
     switch (this.difficulty()) {
-      case Difficulty.EASY:
-        return Difficulty.EASY;
-      case Difficulty.NORMAL:
-        return Difficulty.NORMAL;
+      case 'easy':
+        return 'easy';
+      case 'normal':
+        return 'normal';
```
tomalaforge, line 93: use a mapped type instead of a switch case. Wayyy nicer 😅

PR #667, opened 2024-03-11
Answer:47

BillyOkoth:
Use of union and mapped types

### DIFF
--- a/apps/typescript/enums-vs-union-types/src/app/app.component.ts
+++ b/apps/typescript/enums-vs-union-types/src/app/app.component.ts
@@ -1,14 +1,13 @@
 import { Component, computed, signal } from '@angular/core';
 
-enum Difficulty {
-  EASY = 'easy',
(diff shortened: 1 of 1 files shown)


apps/typescript/enums-vs-union-types/src/app/app.component.ts
```diff
@@ -54,18 +53,17 @@ enum Direction {
...
   readonly difficultyLabel = computed<string>(() => {
     switch (this.difficulty()) {
-      case Difficulty.EASY:
-        return Difficulty.EASY;
-      case Difficulty.NORMAL:
-        return Difficulty.NORMAL;
+      case 'easy':
+        return 'easy';
+      case 'normal':
+        return 'normal';
```
tomalaforge, line 66: you can use a map type to avoid switch case. 💪

svenson95:
For me it look's very good 👍🏼 thanks for submitting your solution 😊

PR #671, opened 2024-03-12
Answer: 38

buhowo:


apps/rxjs/catch-error/src/app/app.component.ts
```diff
@@ -24,38 +24,34 @@ import { Subject, concatMap, map } from 'rxjs';
...
+      {{ (response$ | async) ?? undefined | json }}
     </div>
   `,
   styleUrls: ['./app.component.css'],
 })
 export class AppComponent implements OnInit {
   submit$ = new Subject<void>();
   input = '';
-  response: unknown;
+  response$ = new Observable<string | object>();
```
tomalaforge, line 35: you don't need to instanciate a new observable. you can type it if you want.

PR #672, opened 2024-03-12
Answer:13 angular styling

LKL2017:


apps/angular/styling/src/app/static-text.component.scss
```diff
@@ -0,0 +1,14 @@
...
+
+:host-context(.warning) {
+  font-size: 25px;
+  color: orange;
+}
+
+:host-context(.error) {
+  font-size: 30px;
+  color: red;
+}
```
tomalaforge, line 14: the goal is to use css variables to see the power of this feature and learn how to use it.

LKL2017: Out of my personal programming intuition, I regard this scss file as the entry point to control the `static-text` component style. And do you mean I can define css variables and change them in different selectors? If so, maybe there's only coding style difference in this case, or do I miss something?

tomalaforge: No you can totally use the host-context and do it as you did. 
But I created the challenge to try css variables and see how powerful they are. Because lots of people don't know how to use them.

LKL2017: Cool, got it. Thanks!

PR #680, opened 2024-03-13
Answer:32 change detection bug

LKL2017:


tomalaforge:
using a track is good practise and it's mandatory now, but you are hiding the issue.

You should look at the function call in the template

LKL2017:
In the beginning I found the menu was created every CD by function call, because it returned a new instance. But I didnt realized recreate view and trackBy were two seprate things. I learned from [this comment](https://github.com/tomalaforge/angular-challenges/pull/636#issuecomment-1971440743)  and now I understand why you said 'hide' the issue.
Thanks for your explanation.

PR #688, opened 2024-03-13
style: update button hover effect

svenson95:
As you mentioned in the closed PR, i've added the hover effect again. I made two other changes as well ...

1. Because the opacity hover effect was barely recognizable, i've changed the hover effect of `opacity: 0.9` to `opacity: 0.85`.
2. We have 3 black icon buttons at the top, which have a hover effect with `opacity: 0.66`. In my opinion that's a good idea, because the hover effects on black buttons are even less noticeable. So i've update the hover effect for black buttons to `opacity: 0.66` as well.

Here is one of the 3 black icon buttons which have that 0.66 opacity already.

Bildschirmfoto 2024-03-13 um 17 27 10

---

Do you agree with these changes?

### DIFF
--- a/docs/src/components/SubscriptionForm.astro
+++ b/docs/src/components/SubscriptionForm.astro
@@ -85,10 +85,12 @@ const { data } = await getEntry('i18n', lang ?? 'en');
       text-decoration: none;
       font-size: var(--sl-text-sm);
       background: var(--sl-color-text-accent);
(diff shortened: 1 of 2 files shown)


tomalaforge (approved):

tomalaforge:
thanks a lot. And sorry for the misunderstanding

PR #689, opened 2024-03-13
Answer:34 

lalit-s-rajput:

1. Created two separate components, Input, and list to separate two operation
2. Accommodate OnPush strategy but need to initialize person name array again because changes detection was not able to pick a change in array value (Due to value change and not reference).

### DIFF
--- a/apps/performance/default-onpush/src/app/person-list.component.ts
+++ b/apps/performance/default-onpush/src/app/person-list.component.ts
@@ -1,68 +1,42 @@
-import { Component, Input } from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  Input,
+  OnInit,
+} from '@angular/core';
 
-import { CDFlashingDirective } from '@angular-challenges/shared/directives';
 import { CommonModule } from '@angular/common';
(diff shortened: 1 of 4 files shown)


apps/performance/default-onpush/src/app/person-list.component.ts
```diff
@@ -1,68 +1,42 @@
...
-      this.label = '';
-    }
+  personNames: string[] = [];
+  ngOnInit(): void {
+    this.personNames = this.names;
+  }
+  getPersonName(event: string) {
+    this.personNames = [];
+    this.names?.unshift(event);
+    this.personNames = [...this.names]; // not sure this is the best way. can I use changeDetectoRef instead of this ?
```
tomalaforge, line 40: you should be declarative and you can now use signal inputs 👍

jdegand: Funny enough - a signal implementation can look very similar.   Now that signals are immutable and the spread operator can be used to update a signal.  See this [YouTube video](https://www.youtube.com/watch?v=QOiWLGV9Yrg).

PR #576, opened 2024-02-12
feat(Answer:1): projection

svenson95:
## Checklist for challenge submission

- [x] Start your PR message with Answer:${challenge_number}

PR #695, opened 2024-03-14
Answer: 1

reed756:
Wasn't sure how to get rid of the eslint errors for the any types on the card component but believe I completed all the points including the bonus ones.

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -5,6 +5,7 @@ import { randomCity } from '../../data-access/fake-http.service';
 import { City } from '../../model/city.model';
 import { CardComponent } from '../../ui/card/card.component';
 import { ListItemComponent } from '../../ui/list-item/list-item.component';
+import { ListItemRefDirective } from '../../ui/list-item/list-item.directive';
 
 @Component({
   selector: 'app-city-card',
@@ -12,7 +13,7 @@ import { ListItemComponent } from '../../ui/list-item/list-item.component';
     <app-card [list]="cities()" backgroundColor="rgba(0, 0, 250, 0.1)">
       <img src="assets/img/city.png" width="200px" />
       <ng-template deleteButton let-item>
-        <app-list-item [id]="item.id" (listItemDelete)="deleteCity($event)">
+        <app-list-item (listItemDelete)="deleteCity(item.id)">
           {{ item.name }}
         </app-list-item>
       </ng-template>
@@ -24,7 +25,12 @@ import { ListItemComponent } from '../../ui/list-item/list-item.component';
     </app-card>
   `,
   standalone: true,
-  imports: [CardComponent, NgTemplateOutlet, ListItemComponent],
+  imports: [
+    CardComponent,
+    NgTemplateOutlet,
+    ListItemComponent,
+    ListItemRefDirective,
+  ],
 })
 export class CityCardComponent {
   cities: Signal<City[]> = this.store.cities$;
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,56 @@
+import { NgTemplateOutlet } from '@angular/common';
 import { Component, OnInit } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import {
+  FakeHttpService,
+  randomCity,
+} from '../../data-access/fake-http.service';
+import { CardType } from '../../model/card.model';
+import { City } from '../../model/city.model';
+import { CardComponent } from '../../ui/card/card.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `
+    <app-card
+      [list]="cities"
+      [type]="cardType"
+      customClass="rgba(0, 0, 250, 0.1)">
+      <img src="assets/img/city.png" width="200px" />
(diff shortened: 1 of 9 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,36 @@
...
-  ) {}
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @ContentChild('deleteButton') deleteButton!: TemplateRef<any>;
```
svenson95, line 35: You can check out [my PR](https://github.com/tomalaforge/angular-challenges/pull/576/files/361d3ee8f51f2f8ce6303b98ad665717bc23d038#diff-af224cd0c2986716242c984f2b73acc35291c167462cb08223ec20f1b46eac47R45) to see how to get rid of the `any` type.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,36 @@
...
-      <img
-        *ngIf="type === CardType.TEACHER"
-        src="assets/img/teacher.png"
-        width="200px" />
-      <img
-        *ngIf="type === CardType.STUDENT"
-        src="assets/img/student.webp"
-        width="200px" />
-
+      [ngStyle]="{ 'background-color': customClass }">
```
tomalaforge, line 9: if `customClass` is targeting only the background color, you should rename it for better readibility

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,36 @@
...
-          [type]="type"></app-list-item>
+        @for (item of list; track item.id) {
+          <app-list-item [id]="item.id">
+            <ng-container
+              *ngTemplateOutlet="
+                deleteButton;
+                context: {
+                  $implicit: item.id,
+                  name: item.name,
+                  firstName: item.firstName
```
tomalaforge, line 20: why are you not setting the full object ? 
Because you don't know the property of the inputed list. It can be anything if you want your component to be highly reusable

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,36 @@
...
-      </button>
+      <ng-content select="button"></ng-content>
     </div>
   `,
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
+  imports: [ListItemComponent, NgTemplateOutlet, NgStyle],
 })
 export class CardComponent {
   @Input() list: any[] | null = null;
```
tomalaforge, line 33: don't use any, prefer generics

apps/angular/projection/src/app/ui/card/card.component.ts
tomalaforge, line 35: prefer directive over magic string.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,52 @@
...
 
-  ngOnInit(): void {}
+  constructor(
+    private http: FakeHttpService,
+    private store: CityStore,
+  ) {}
+
+  ngOnInit(): void {
+    this.http.fetchCities$.subscribe((c) => this.store.addAll(c));
+    this.store.cities$.subscribe((c) => this.cities.set(c));
```
tomalaforge, line 42: prefer reactivity over impertivity. 
you can use signal as well

apps/angular/projection/src/app/ui/card/card.component.ts
reed756: Thanks for your comment and sharing your code, I didn't think to apply a directive but have read the reasoning behind it and implemented it in my latest commit.

reed756:
@tomalaforge Thanks for your review I have tried to apply all the comments you've made in my latest commit, please let me know what you think.

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -17,19 +14,9 @@ import { CardType } from '../../model/card.model';
...
-  ) {}
-
-  delete(id: number) {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.deleteOne(id);
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.deleteOne(id);
-    }
+  deleteId() {
+    this.listItemDelete.emit(this.id);
```
tomalaforge, line 20: here you don't need the id. this can be a simple event Emitter that trigger the delete action and you pass the id in the parent.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,30 @@
...
-  ) {}
-
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Input() list: { id: number }[] | null = null;
```
tomalaforge, line 26: it's better to use generic

```
export class CardComponent<T extends {id: number}> {
  @Input() list: T[] | null = null;
  ```

tomalaforge:
Yes this is way bettter. great job

reed756:
Thanks for the review @tomalaforge

PR #696, opened 2024-03-16
docs: add note to bug-cd-md

jdegand:
Don't know if you want to add or spoil "trackBy" as possible fix so most people don't miss the lesson of this challenge.  I suggest you downvote my bug-cd answers, so people recognize that isn't the correct solution. 

I am not a fan of this line,`If you comment out this line, the application loads again, however the bug is not inside the Angular Framework. 😅😯`, in the hint, as it can be potentially misleading.  The bug is related to how `*ngFor` operates.

### DIFF
--- a/docs/src/content/docs/challenges/angular/32-bug-cd.md
+++ b/docs/src/content/docs/challenges/angular/32-bug-cd.md
@@ -25,6 +25,10 @@ In this small application, we have a navigation menu to route our application to
 
 The goal of the challenge is to debug this application and make it work.
 
+:::note
+Without knowing the exact reason for the issue, you can "fix" the error and get the program to function. One such approach would be to memoize the `getMenu` function. The application might function, but you will miss the lesson of this challenge. Make sure you can explain why your fix works.  
+:::
+
 ## Hints
 
 <details>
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/challenges/angular/32-bug-cd.md
```diff
@@ -25,6 +25,10 @@ In this small application, we have a navigation menu to route our application to
 
 The goal of the challenge is to debug this application and make it work.
 
+:::note
+Without knowing the exact reason for the issue, you can "fix" the error and get the program to function. One such approach would be to memoize the `getMenu` function. The application might function, but you will miss the lesson of this challenge. Make sure you can explain why your fix works.
```
svenson95, line 29: Really good idea! I'd improve the wording just a little bit. My suggestion is changing this last sentence to something like that ...

> The application might function, but you will miss the lesson of this challenge. Make sure you can explain why your fix works.  

To ...

> The application might work again, but make sure you really understand the problem and it's consequences. To make it work isn't always enough, fixing this bug in a wrong way can cause a loss of performance - or may lead to other problems that originate from an incorrect implementation.

jdegand: Great suggestion. I cleaned up the grammar slightly and tweaked the hint.  I think it flows better now.

jdegand: Looking at it again, `function` was repeated too much so I removed it from the memoize sentence.

tomalaforge:
i like what you add to the documentation.

But i just want to react to your comment.
If you add just the `trackBy` function, you are hiding the issue. This will work but your list will always get rerender at each change detection cycle.

tomalaforge (approved):

PR #697, opened 2024-03-16
fix(docs): missing projection challenge solution buttons

svenson95:
I've noticed that in the first challenge "Projection" the solution buttons are not displayed (Community Solutions & Author's Solution). Close this PR if that's on purpose, if not, here's the fix to show them.


tomalaforge (approved):

tomalaforge:
Indeed thanks a lot

PR #698, opened 2024-03-17
Answer:35 (create memorization pipe)

lalit-s-rajput:

feat: created memoization pipe
image

### DIFF
--- a/apps/performance/memoized/src/app/fib.pipe.ts
+++ b/apps/performance/memoized/src/app/fib.pipe.ts
@@ -1,0 +1,28 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+const fibonacci = (num: number): number => {
+  if (num === 1 || num === 2) {
+    return 1;
+  }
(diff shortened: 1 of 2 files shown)


apps/performance/memoized/src/app/fib.pipe.ts
```diff
@@ -0,0 +1,28 @@
...
+   * @param {}
+   * @param fibObject
+   * Cant create fibObject here because on every pipe call it is creating new object instance so creating from person-list
+   * and sending from there.
+   */
+  transform(value: number, fibObject: any) {
+    if (fibObject[value]) {
+      return fibObject[value];
+    }
+    fibObject[value] = fibonacci(value);
```
tomalaforge, line 25: What is the goal of this fibObject ?

lalit-s-rajput: I did not wanted to calculate Fibonacci of same numbers so stored them in object to reuse so that if number appears again it will get from object and will not calculate.

PR #701, opened 2024-03-20
Answer 1: finishing projection

michal13171:

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -13,13 +13,6 @@ import { CardComponent } from '../../ui/card/card.component';
       [type]="cardType"
       customClass="bg-light-red"></app-card>
   `,
-  styles: [
-    `
-      .bg-light-red {
-        background-color: rgba(250, 0, 0, 0.1);
-      }
-    `,
-  ],
   standalone: true,
   imports: [CardComponent],
 })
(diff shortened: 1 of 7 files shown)


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,40 @@
...
-  template: 'TODO City',
+  template: `
+    <app-card
+      [list]="cities"
+      [type]="cardType"
+      customClass="bg-light-red"></app-card>
+  `,
+  styles: [
+    `
+      ::ng-deep .bg-light-red {
```
svenson95, line 18: Avoid using `::ng-deep`, you don't need it to solve this challenge. If you struggle with it, check out the community solutions or authors solution.

tomalaforge:
You need to remove the logic from cardComponent.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -40,22 +43,15 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
   @Input() type!: CardType;
   @Input() customClass = '';
 
   CardType = CardType;
 
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore,
-  ) {}
+  constructor(private cardService: CardService) {}
```
tomalaforge, line 52: nice try, but the goal is to create a dumb component. use `ng-content`, `ng-template-outlet` to project content from the parent

CardType should be deleted at the end

PR #708, opened 2024-03-23
docs: faq note

jdegand:
Added a note about having to run install if `npm ci` fails.  I think there is a giscus comment about this.   If you run `npm ci` after a `git pull`, you will need to install first.

### DIFF
--- a/docs/src/content/docs/guides/faq.md
+++ b/docs/src/content/docs/guides/faq.md
@@ -11,7 +11,7 @@ sidebar:
 <details>
   <summary>Why is my application not starting, or why do I encounter errors in my terminal when I run `nx serve`?</summary>
   
-  Most of the time, this issue arises because your node_modules are outdated, and you need to update them by running `npm ci`.
+  Most of the time, this issue arises because your node_modules are outdated, and you need to update them by running `npm ci`.  You might need to run `npm i` first if `npm ci` fails.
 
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/guides/faq.md
```diff
@@ -11,7 +11,7 @@ sidebar:
 <details>
   <summary>Why is my application not starting, or why do I encounter errors in my terminal when I run `nx serve`?</summary>
   
-  Most of the time, this issue arises because your node_modules are outdated, and you need to update them by running `npm ci`.
+  Most of the time, this issue arises because your node_modules are outdated, and you need to update them by running `npm ci`.  You might need to run `npm i` first if `npm ci` fails.
```
tomalaforge, line 14: No, you should delete your node module if npm ci fails.

Npm i will reinstall the package and create a new lock file. We don't want that. Most of the time it's because of a mismatch version between your current node module and the new node module you want to install.

jdegand: I tried to keep the note vague to apply to multiple scenarios.

I had an issue where the lock file was different after a `git pull`.  In that scenario, you can revert the lock file or `npm i`.

jdegand: If you install other packages on other branches, you can have a package lock issue after `git pull`  (Didn't think of this at the time) .

PR #710, opened 2024-03-27
Answer: 21

Sanjar1304:
Anchor Navigation task is done with ViewportScroller methods


tomalaforge:
one challenge per PR

Sanjar1304:
> one challenge per PR

Ok. I understand. For each task I will create separate branch and I will open per PR only for those branchs separately.

PR #711, opened 2024-03-28
feat: challenge 48 Pipe alternative

svenson95:
Here is a new proposal for a challenge. For me it's a interesting topic and after the pipe challenges, i delved a little bit deeper into the topic. So i found this article about [It’s ok to use function calls in Angular templates](https://itnext.io/its-ok-to-use-function-calls-in-angular-templates-ffdd12b0789e) and tried it out myself. In my opinion this is a nice extension to the `@Pipe()` challenges.

This is a first draft, so be encouraged to suppose changes or ideas for the challenge project. My initial idea is to prepare a function - to format a date - which is called in the template. The goal of the challenge would be refactor this function with a `memo` function. 

I'm also thinking about whether I should somehow display in the UI how often the function or pipe is called - before and after. For example, by creating an element each time the function is called and putting it into an array. To visualize the improvement.


tomalaforge:
Hey @svenson95, I can see you are back to a new challenge. 

I'll look more in-depth tomorrow but warning, memo is useless now that computed is present. So I won't create a challenge on this utility function. The article of enea is great but deprecated

svenson95:
Okay, we could also mention that since v17 with computed this memo approach is unnecessary. But maybe for some people, who have to work with angular below v17, this can be useful to know.

jdegand:
This [article](https://medium.com/@ineedsomemeat/angular-optimization-memoized-pipe-functions-in-templates-75f62e16df5a) and this [article](https://medium.com/@bansal.suneet/memo-decorator-with-angular-pipe-big-performance-boost-73f94b5c728a) could be of interest.  

Few thoughts if you want to rework the concept of the challenge:
- You could shift to an `Impure pipe` challenge.  See this [article](https://angularindepth.com/posts/1061/the-essential-difference-between-pure-and-impure-pipes-in-angular-and-why-that-matters).  Problem would be that there really isn't much to cover.   
-  If you want to shift the focus to `computed`, a challenge could focus on conditionals inside `computed` functions.  If you have a ternary with multiple signals, unexpected results can occur.  Or even an `effect` vs `computed` challenge.

tomalaforge:
I'm sorry @svenson95, but I don't see the benefits of this challenge. There is no real added value and people should focus on using Pipe instead of useMemo. 
useMemo has side effects if used multiple times inside a template and you will not lose a lot of performance by just calling the function inside your template. 
I would rather focus on creating challenges about signal, and all signal inputs, outputs, and so on. 
Signal is very new and lots of people struggle with it. 

I like the idea of @jdegand about the computed function with a ternary inside. Or maybe create a challenge where we need to use the  untrack function. 

What do you think?

svenson95:
No problem at all. That's why i first started only the article but not the challenge. I wasn't sure if that's a good challenge, it was just interesting for me. Challenges about signals sounds more interesting though, so i'll work on that. This PR can be closed, i will open up a new one.

PR #713, opened 2024-03-29
Answer:1

darkv314:
Thanks a lot for creating this project, I'm learning a lot while having fun <3

### DIFF
--- a/apps/angular/projection/src/app/ui/card/card.component.ts
+++ b/apps/angular/projection/src/app/ui/card/card.component.ts
@@ -1,61 +1,45 @@
-import { NgFor, NgIf } from '@angular/common';
-import { Component, Input } from '@angular/core';
-import { randStudent, randTeacher } from '../../data-access/fake-http.service';
-import { StudentStore } from '../../data-access/student.store';
-import { TeacherStore } from '../../data-access/teacher.store';
-import { CardType } from '../../model/card.model';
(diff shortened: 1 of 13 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,45 @@
...
+  },
+  imports: [NgIf, NgFor, ListItemComponent, NgTemplateOutlet, RowRefDirective],
+})
+export class CardComponent<T extends { id: number }> {
+  @Input() list: T[] | null = null;
+  @ContentChild(RowRefDirective, { read: TemplateRef })
+  rowTemplate!: TemplateRef<{
+    $implicit: T;
+  }>;
+  @Output() addItem = new EventEmitter<void>();
```
tomalaforge, line 44: you can use signal queries now

tomalaforge:
nice job, but you can go one step further with all the new signal features

apps/angular/projection/src/app/ui/card/card.component.ts
darkv314: Sure! I just changed that, I'm not sure if I did it correctly tho :( If there's another way, please let me know.

tomalaforge: YEs that's great, you can do it for input and output as well.

PR #716, opened 2024-03-31
Answer:6

dpinalopez:

### DIFF
--- a/apps/angular/permissions/src/app/authGuard.guard.ts
+++ b/apps/angular/permissions/src/app/authGuard.guard.ts
@@ -1,6 +1,44 @@
-import { type CanActivateFn } from '@angular/router';
+import { inject } from '@angular/core';
+import { Router, type CanActivateFn } from '@angular/router';
+import { UserStore } from './user.store';
 
 export const authGuardGuard: CanActivateFn = (route, state) => {
   //
+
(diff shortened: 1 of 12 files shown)


apps/angular/permissions/src/app/authGuard.guard.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (
+    loggedUserWeight.weight < pathWeight.weight ||
+    loggedUser.name === currentRoute
+  ) {
+    return true;
+  } else {
+    router.navigateByUrl(`/${loggedUserWeight.route}`);
+    return false;
+  }
+};
```
tomalaforge, line 35: I think you made things over complicated,

guards are functions. You can pass parameters to them. Just pass it the list of role allowed for this specific route and check if that match the user

apps/angular/permissions/src/app/user.model.ts
```diff
@@ -47,3 +52,45 @@ export const everyone: User = {
...
+export interface RoleWeight {
+  name: string;
+  route: string;
+  weight: number;
+}
+export const RolesWeight: RoleWeight[] = [
+  {
+    name: 'enter',
+    route: 'enter',
+    weight: 0,
```
tomalaforge, line 64: what is the weight for ?

PR #725, opened 2024-04-02
Answer 1 # refactor: improve card component for reusability

sefatanam:
In the current state, the application has a functional dashboard with teacher and student cards. However, the developer experience is suboptimal due to the need for modifications in card.component.ts every time a new card is implemented. This is not scalable, especially for shared components across multiple projects.

To address this, I've refactored the code to introduce a CardComponent that can be customized without altering the core component. This improvement enhances reusability and maintainability. With this refactor in place, I'm now ready to proceed with implementing the CityCardComponent without any need to touch the CardComponent.

### DIFF
--- a/apps/angular/projection/src/app/component/student-card/student-card.component.ts
+++ b/apps/angular/projection/src/app/component/student-card/student-card.component.ts
@@ -1,40 +1,62 @@
-import { Component, OnInit } from '@angular/core';
-import { FakeHttpService } from '../../data-access/fake-http.service';
+import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { FakeHttpService, randStudent } from '../../data-access/fake-http.service';
 import { StudentStore } from '../../data-access/student.store';
-import { CardType } from '../../model/card.model';
 import { Student } from '../../model/student.model';
 import { CardComponent } from '../../ui/card/card.component';
+import { AsyncPipe } from '@angular/common';
+import { Observable } from 'rxjs';
 
 @Component({
   selector: 'app-student-card',
   template: `
     <app-card
-      [list]="students"
-      [type]="cardType"
-      customClass="bg-light-green"></app-card>
+      [list]="students$ | async"
+      [customClass]="'bg-light-green'"
+      [listTemplateRef]="listTemplateRef"
+      (delete)="delete($event)">
+   
+      <img cardImage
+        src="assets/img/student.webp"
+        width="200px" /> 
+
+        <button cardActionButton
+        class="rounded-sm border border-blue-500 bg-blue-300 p-2"
+        (click)="addNewStudent()">
+        Add
+      </button>
+    </app-card>
+
+    <ng-template #listTemplateRef let-item>
+        <p>{{item.firstName}}</p>
+    </ng-template>
   `,
   standalone: true,
   styles: [
     `
-      ::ng-deep .bg-light-green {
+      .bg-light-green {
         background-color: rgba(0, 250, 0, 0.1);
       }
     `,
   ],
-  imports: [CardComponent],
+  imports: [CardComponent, AsyncPipe],
+  encapsulation: ViewEncapsulation.None
 })
 export class StudentCardComponent implements OnInit {
-  students: Student[] = [];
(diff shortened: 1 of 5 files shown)


apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,34 @@
...
-
+      class="flex w-fit flex-col gap-3 rounded-md border-2 border-black p-4" [class]="customClass">
+        <ng-content select="[cardImage]"></ng-content>
       <section>
-        <app-list-item
-          *ngFor="let item of list"
-          [name]="item.firstName"
-          [id]="item.id"
-          [type]="type"></app-list-item>
+        @for (item of list; track $index) {
```
tomalaforge, line 12: it's better to track on a unique property is possible.
if you change the list, the $index is not optimal since all index will change so the list will still get refresh entirely

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,34 @@
...
-
-  constructor(
-    private teacherStore: TeacherStore,
-    private studentStore: StudentStore,
-  ) {}
+export class CardComponent<T extends {id:number}> {
+  @Input() list: T[] | null = null;
+  @Input() customClass !:string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  @Input() listTemplateRef: TemplateRef<any> | null = null;
```
tomalaforge, line 31: prefer to use contentchild to get template from the parent.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,34 @@
...
 
-  addNewItem() {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.addOne(randTeacher());
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.addOne(randStudent());
-    }
-  }
+  @Output() delete: EventEmitter<number> = new EventEmitter();
```
tomalaforge, line 33: you can use the new set of signal queries

apps/angular/projection/src/app/ui/list-item/list-item.component.ts
```diff
@@ -17,19 +14,5 @@ import { CardType } from '../../model/card.model';
...
-  ) {}
-
-  delete(id: number) {
-    if (this.type === CardType.TEACHER) {
-      this.teacherStore.deleteOne(id);
-    } else if (this.type === CardType.STUDENT) {
-      this.studentStore.deleteOne(id);
-    }
-  }
+  @Output() delete : EventEmitter<number> = new EventEmitter();
```
tomalaforge, line 17: id is not needed here, delete is only a event emitter

apps/angular/projection/src/app/component/student-card/student-card.component.ts
```diff
@@ -1,40 +1,62 @@
...
     `,
   ],
-  imports: [CardComponent],
+  imports: [CardComponent, AsyncPipe],
+  encapsulation: ViewEncapsulation.None
 })
 export class StudentCardComponent implements OnInit {
-  students: Student[] = [];
-  cardType = CardType.STUDENT;
+  students$: Observable<Student[]> = this.store.students$;
```
tomalaforge, line 45: prefer to use signal to bind with the tempalte.

tomalaforge:
good job and nice description

apps/angular/projection/src/app/ui/card/card.component.ts
sefatanam: now I updated this line by `  @for (item of list; track item.id) {`, I hope now it track the changes in better way

apps/angular/projection/src/app/ui/card/card.component.ts
sefatanam: Thank you for your suggestion. I appreciate your input and will certainly explore the possibility of replacing this LOC  with signal queries. At the moment, I'm not fully acquainted with this approach, but I'm committed to learning and implementing best practices.

PR #726, opened 2024-04-02
Answer: 1-projection

tsironis13:
- Fix projection challenge by converting card and item components to presentational ones.

- Card and item are now library components.

### DIFF
--- a/apps/angular/projection/src/app/ui/card/card.component.html
+++ b/apps/angular/projection/src/app/ui/card/card.component.html
@@ -1,0 +1,35 @@
+<ng-container [ngTemplateOutlet]="cardWrapper"></ng-container>
+<ng-template #cardWrapper>
+  <div
+    class="flex w-fit flex-col gap-3 rounded-md border-2 border-black p-4"
+    [class]="customClass()">
+    <ng-container
+      [ngTemplateOutlet]="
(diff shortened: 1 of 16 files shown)


apps/angular/projection/src/app/ui/card/card.component.html
```diff
@@ -0,0 +1,35 @@
...
+      This is the default card header
+    </ng-template>
+
+    <section>
+      @for (item of list(); track item.id) {
+        <app-list-item [item]="item" (onDeleteItem)="deleteItem($event)">
+          <ng-template #itemContentTemplate>
+            <div>{{ item[prop()] }}</div>
+          </ng-template>
+        </app-list-item>
```
tomalaforge, line 20: you made things a bit complicated and you coupled listItem with card, So you cannot use something else inside your card. But it's working.

tomalaforge:
Your solution is working but overcomplicated. You should look at other solution to see how you can simplify your work and make is more maintainable and way for customizable. 

great work

apps/angular/projection/src/app/ui/card/card.component.html
tsironis13: I get your point, I will refactor and decouple the card component. Thanks for your instant feedback.

PR #727, opened 2024-04-02
Answer:9

dpinalopez:


tomalaforge:
For this specific use case your pipe works but the goal is to wrap any functions. Your pipe will only work for this specific use case.

PR #730, opened 2024-04-03
Answer1: Projection challenge

sunilbaba:


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,54 @@
...
+      </ng-template>
+    </app-card>
+  `,
   standalone: true,
-  imports: [],
+  imports: [CardComponent, ListItemComponent],
 })
 export class CityCardComponent implements OnInit {
-  constructor() {}
+  cities: City[] = [];
```
svenson95, line 35: I'd prefer using a signal in here, something like `cities = toSignal(this.store.cities$)`. You could refactor the `cities$` subject to a signal as well.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,54 @@
...
+    <app-card
+      [list]="cities"
+      [type]="cardType"
+      (add)="addNew()"
+      customClass="bg-light-red">
+      <img
+        src="assets/img/student.webp"
+        alt="Placeholder for the student card"
+        width="200px" />
+      <ng-template #rowRef let-city>
```
svenson95, line 24: #rowRef is a magic string, try to [avoid this](https://softwareengineering.stackexchange.com/a/365344).

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -12,50 +16,45 @@ import { ListItemComponent } from '../list-item/list-item.component';
...
+        .bg-light-green {
+          background-color: rgba(250, 0, 0, 0.1);
+        }
+      }
+    `,
+  ],
+  imports: [NgFor, ListItemComponent, NgTemplateOutlet],
 })
 export class CardComponent {
   @Input() list: any[] | null = null;
```
svenson95, line 51: We should'nt use `any[]` type right here, type `any` is kind of "i dont care about the type" - that destroys the concept of **type**script.

sunilbaba: Hi @svenson95 ,

 Students component sent Students Array list under the list property. Same with teachers and city card components as well. I don't want my card component to have typing as **Student | Card | Teacher**
 
  I want to avoid the scenario of adding new types inside card component every time there is a new  implementation layer added for card component into the application.
  
  How to add this scenario ? can we have dynamic typing for better type casting ?

apps/angular/projection/src/app/component/city-card/city-card.component.ts
sunilbaba: Hi @svenson95 ,

 How to get rid of this magic string ? Any examples on how to avoid this in content projection case ?

svenson95: Yes you just need to create a `Directive`, [here](https://medium.com/@thomas.laforge/ngtemplateoutlet-type-checking-5d2dcb07a2c6) is a nice article from @tomalaforge about this topic

apps/angular/projection/src/app/ui/card/card.component.ts
svenson95: Yeah it's possible to use a generic `T` type instead of `any`, so you can pass different types to this `list` input. Furthermore you can take advantage of `ngTemplateContextGuard`, but this is part of another challenge - [Typed ContextOutlet](https://angular-challenges-620x3ppdp-tomalaforges-projects.vercel.app/challenges/angular/4-context-outlet-typed/). I'd recommend you to checkout the community & author solutions of this challenge as well, just to see different ways of solving this problem and maybe learn something new 🚀

PR #733, opened 2024-04-05
Answer:1

proselin:

## Note
Thanks for making this challenges pages :3


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,55 @@
...
+
+  addOne() {
+    this.store.addOne(randomCity());
+  }
+
+  removeOne(id: number) {
+    this.store.deleteOne(id);
+  }
+
+  protected readonly of = of;
```
svenson95, line 54: i think that of is unused?

PR #734, opened 2024-04-05
Answer:21

Sanjar1304:
Following your advice I managed and learned how to make PR for per task. Thanks so much Thomas !!!

### DIFF
--- a/apps/angular/anchor-scrolling/src/app/foo.component.ts
+++ b/apps/angular/anchor-scrolling/src/app/foo.component.ts
@@ -1,5 +1,4 @@
-import { ViewportScroller } from '@angular/common';
-import { Component, inject } from '@angular/core';
+import { Component } from '@angular/core';
 import { RouterModule } from '@angular/router';
 import { NavButtonComponent } from './nav-button.component';
 
@@ -33,8 +32,6 @@ import { NavButtonComponent } from './nav-button.component';
   `,
 })
 export class FooComponent {
-  scroller = inject(ViewportScroller);
-
   scrollTo(anchor: string) {
     const element = document.getElementById(anchor);
     if (element) {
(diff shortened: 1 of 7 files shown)


apps/angular/anchor-scrolling/src/app/foo.component.ts
```diff
@@ -1,15 +1,44 @@
...
+      section 2
+      <button class="m-2 block w-fit rounded-md border border-red-500 p-4">
+        Section 1
+      </button>
+    </div>
   `,
 })
-export class FooComponent {}
+export class FooComponent {
+  scroller = inject(ViewportScroller);
```
svenson95, line 36: `scroller` is unsed?

apps/angular/anchor-scrolling/src/app/home.component.ts
```diff
@@ -1,20 +1,34 @@
...
-      <nav-button href="#top">Scroll Top</nav-button>
+      <nav-button>Scroll Top</nav-button>
     </div>
   `,
 })
-export class HomeComponent {}
+export class HomeComponent {
+  scroller = inject(ViewportScroller);
+
+  scrollTo(anchor: string) {
```
svenson95, line 27: You have the same code for `scrollTo` in `FooComponent`, you should refactor this in a service for example.

svenson95:
Good job 👍 but i'd recommend you to checkout the authors solutions as well, there is a way to add this functionality without writing any code in the class.

apps/angular/anchor-scrolling/src/app/foo.component.ts
Sanjar1304: removed unused scroller

apps/angular/anchor-scrolling/src/app/home.component.ts
Sanjar1304: here also removed unused scroller

Sanjar1304: Add service and scrolling function moved to service file

PR #737, opened 2024-04-05
Answer:8

Sanjar1304:
added pipe and removed method

### DIFF
--- a/apps/angular/pipe-easy/src/app/app.component.ts
+++ b/apps/angular/pipe-easy/src/app/app.component.ts
@@ -1,21 +1,21 @@
-import { NgFor } from '@angular/common';
+import { NgFor, TitleCasePipe } from '@angular/common';
 import { Component } from '@angular/core';
+import { ComputationPipe } from './computation.pipe';
 
 @Component({
   standalone: true,
-  imports: [NgFor],
+  imports: [NgFor, ComputationPipe, TitleCasePipe],
   selector: 'app-root',
   template: `
-    <div *ngFor="let person of persons; let index = index">
-      {{ heavyComputation(person, index) }}
-    </div>
+    @for (person of persons; track persons; let index = $index) {
+      <div>
+        {{ person | computation: index | titlecase }}
+      </div>
+    } @empty {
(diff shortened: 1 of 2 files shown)


apps/angular/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,21 @@
-import { NgFor } from '@angular/common';
+import { NgFor, TitleCasePipe } from '@angular/common';
 import { Component } from '@angular/core';
+import { ComputationPipe } from './computation.pipe';
 
 @Component({
   standalone: true,
-  imports: [NgFor],
+  imports: [NgFor, ComputationPipe, TitleCasePipe],
```
tomalaforge, line 7: no more ngFor needed

apps/angular/pipe-easy/src/app/app.component.ts
```diff
@@ -1,21 +1,21 @@
...
   selector: 'app-root',
   template: `
-    <div *ngFor="let person of persons; let index = index">
-      {{ heavyComputation(person, index) }}
-    </div>
+    @for (person of persons; track persons; let index = $index) {
+      <div>
+        {{ person | computation: index | titlecase }}
```
tomalaforge, line 12: you could now use a computed function 😅

apps/angular/pipe-easy/src/app/app.component.ts
Sanjar1304: Removed ngFor from imports array.  I forgot to delete ngfor from there. New syntax of ngFor does not require to import into component imports array. That's cool thing.

apps/angular/pipe-easy/src/app/app.component.ts
Sanjar1304: Should I convert my custom pipe to funtion ??? I did not understand a little bit.

tomalaforge: No no, it's just an alternative. but pipes are perfect here

PR #741, opened 2024-04-07
Answer:1

annrobles:

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,42 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, ViewEncapsulation } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import { FakeHttpService } from '../../data-access/fake-http.service';
+import { CardType } from '../../model/card.model';
+import { City } from '../../model/city.model';
+import { CardComponent } from '../../ui/card/card.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `
+    <app-card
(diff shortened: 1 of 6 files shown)


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,42 @@
...
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `
+    <app-card
+      [list]="cities"
+      [type]="cardType"
+      [imageSrc]="'assets/img/city.png'"
+      customClass="bg-light-blue"></app-card>
+  `,
+  encapsulation: ViewEncapsulation.None,
```
svenson95, line 17: FYI: it's dangerous to set `ViewEncapsulation.None`. In this case right here, the styling for `bg-light-blue` is set globally. So if you'd have the same css class in a different component, this styles would be applied.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,42 @@
...
-  ngOnInit(): void {}
+  constructor(
+    private http: FakeHttpService,
+    private store: CityStore,
+  ) {}
+
+  ngOnInit(): void {
+    this.http.fetchCities$.subscribe((cities) => this.store.addAll(cities));
+
+    this.store.cities$.subscribe((cities) => (this.cities = cities));
```
svenson95, line 40: no need to subscribe two times, you could use a signal for `cities` in the `CityStore`, and here in line 29 just use it like `cities = this.store.cities;`

svenson95:
Good job 👍 you could improve your code, checkout the community & author solutions as well. For example the statement `if (this.type === CardType.TEACHER) { ... }` is redundant in your code.

PR #756, opened 2024-04-09
Answer: 6

alcaidio:

Signal and RXJS in the directive ! 😁

### DIFF
--- a/apps/angular/permissions/src/app/role.directive.ts
+++ b/apps/angular/permissions/src/app/role.directive.ts
@@ -1,0 +1,57 @@
+import {
+  Directive,
+  Input,
+  TemplateRef,
+  ViewContainerRef,
+  booleanAttribute,
+  inject,
+} from '@angular/core';
+import { takeUntilDestroyed, toObservable } from '@angular/core/rxjs-interop';
+import { filter, tap } from 'rxjs';
+import { Role } from './user.model';
+import { UserStore } from './user.store';
+
+@Directive({ selector: '[appHasRole]', standalone: true })
+export class HasRoleDirective {
+  @Input({ required: true }) appHasRole!: Role[];
+
+  private templateRef = inject(TemplateRef);
+  private viewContainer = inject(ViewContainerRef);
+  private user$ = toObservable(inject(UserStore).user);
+
+  constructor() {
+    this.user$
+      .pipe(
+        tap(() => this.viewContainer.clear()),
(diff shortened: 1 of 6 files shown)


apps/angular/permissions/src/app/user.model.ts
```diff
@@ -1,5 +1,9 @@
-export type Role = 'MANAGER' | 'WRITER' | 'READER' | 'CLIENT';
-
+export enum Role {
+  MANAGER = 'MANAGER',
+  WRITER = 'WRITER',
+  READER = 'READER',
+  CLIENT = 'CLIENT',
+}
```
tomalaforge, line 6: you should prefer union type over enums. That's a question of preference, but at least you don't have to decalre the enum inside your template

apps/angular/permissions/src/app/user.store.ts
```diff
@@ -1,15 +1,14 @@
...
+import { Injectable, signal } from '@angular/core';
 import { User } from './user.model';
 
 @Injectable({
   providedIn: 'root',
 })
 export class UserStore {
-  private user = new BehaviorSubject<User | undefined>(undefined);
-  user$ = this.user.asObservable();
+  private _user = signal<User | undefined>(undefined);
```
tomalaforge, line 8: you can now use # if you want a real private property

apps/angular/permissions/src/app/role.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+  inject,
+} from '@angular/core';
+import { takeUntilDestroyed, toObservable } from '@angular/core/rxjs-interop';
+import { filter, tap } from 'rxjs';
+import { Role } from './user.model';
+import { UserStore } from './user.store';
+
+@Directive({ selector: '[appHasRole]', standalone: true })
+export class HasRoleDirective {
+  @Input({ required: true }) appHasRole!: Role[];
```
tomalaforge, line 16: you can use signal input

apps/angular/permissions/src/app/role.directive.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  private templateRef = inject(TemplateRef);
+  private viewContainer = inject(ViewContainerRef);
+  private user$ = toObservable(inject(UserStore).user);
```
tomalaforge, line 20: you could try fully signals, with computed and effect

PR #761, opened 2024-04-11
fix: create-challenge-typos & extra clarification

jdegand:
I saw some typos and did some tweaks. Instead of making 2 PRs, I combined it here.  I added a line about using an extension to generate a challenge.   Forms were missing from the list of challenges here as well.


tomalaforge:
Thanks but "forms" is also missing inside the schematics schema.

tomalaforge (approved):

PR #762, opened 2024-04-11
Answer:1

devXodia:
this is my solution for challenge 1. 

would appreciate feedback since im in my first job hunting phase. 

i didn't use signals or directives so far. I have to get a better understanding on how to use it and what it really does before...so i diddn't use it here.

### DIFF
--- a/apps/angular/projection/src/app/component/city-card/city-card.component.ts
+++ b/apps/angular/projection/src/app/component/city-card/city-card.component.ts
@@ -1,13 +1,32 @@
 import { Component, OnInit } from '@angular/core';
+import { CityStore } from '../../data-access/city.store';
+import { FakeHttpService } from '../../data-access/fake-http.service';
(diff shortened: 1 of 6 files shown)


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,36 @@
...
+  ) {}
 
-  ngOnInit(): void {}
+  cities: City[] = [];
+  cardType = CardType.CITY;
+
+  ngOnInit(): void {
+    this.http.fetchCities$.subscribe((s) => this.cityStore.addAll(s));
+
+    this.cityStore.cities$.subscribe((s) => (this.cities = s));
```
svenson95, line 34: You don't need to subscribe two times, checkout [my solution](https://github.com/tomalaforge/angular-challenges/pull/576/files) to see a different way to get the data.

PR #766, opened 2024-04-12
Answer: 5 - with tanStack query 🚀

alcaidio:

## Warning:

  - $5 per review
  - $25 for lifetime reviews

Alternatively, you can still submit your PR to join the list of answered challenges.

**Oh no, that's new :( That means you'll never review all my PR now... @tomalaforge**


tomalaforge:
No I need to add a new line, for challenge creators, I will review them. But I'm bit burn out of PRs. It's taking way too much of my time and I cannot keep up. I cannot find time to create new challenges.

apps/angular/crud/src/app/app.config.ts
```diff
@@ -1,5 +1,13 @@
...
+import { ApplicationConfig } from '@angular/core';
+import {
+  QueryClient,
+  provideAngularQuery,
+} from '@tanstack/angular-query-experimental';
+
 export const appConfig: ApplicationConfig = {
-  providers: [importProvidersFrom(HttpClientModule)],
+  providers: [
+    provideHttpClient(withFetch()),
```
tomalaforge, line 10: Why do you use withFetch ?

apps/angular/crud/src/app/item.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  todo = input.required<Todo>();
+
+  #api = inject(TodosApiService);
+
+  update = injectMutation((client) => ({
+    mutationFn: (todo: Todo) => lastValueFrom(this.#api.updateTodo$(todo)),
+    onSuccess: () => {
+      client.invalidateQueries({ queryKey: ['todos'] });
+    },
+  }));
```
tomalaforge, line 50: I like to create another file called 'todo.query.ts' where I put all my related queries. it makes the code nicer.

tomalaforge: I also like to create a factory function for my keys.

tomalaforge:
Tanstack Query is 😍

alcaidio:
> No I need to add a new line, for challenge creators, I will review them. But I'm bit burn out of PRs. It's taking way too much of my time and I cannot keep up. I cannot find time to create new challenges.

Yes I understand, do as you can. I think you have to get the community to review itself because it's not sustainable like that. I look forward to hearing from you. Maybe launch an RFC to find solutions to this?

alcaidio:
> Tanstack Query is 😍

It's thanks to you that I discovered this. I fell in love with it the second I saw it 😁

apps/angular/crud/src/app/item.component.ts
alcaidio: Yes good idea, factory function ? Why not an enum ?

apps/angular/crud/src/app/app.config.ts
tomalaforge: it's mandatory in SSR

apps/angular/crud/src/app/item.component.ts
tomalaforge: First I hate enums 😅, 
But then for your case, why not because todos is static, but then, you will need to add variables to create your key
```
queryKey: ['todos', todoId]
````

svenson95:
> No I need to add a new line, for challenge creators, I will review them. But I'm bit burn out of PRs. It's taking way too much of my time and I cannot keep up. I cannot find time to create new challenges.

i think that solution with 5€ per review and the lifetime option is really good 👍 so that time and work you put into this project pays off, you deserve this 🚀

alcaidio:
> i think that solution with 5€ per review and the lifetime option is really good 👍 so that time and work you put into this project pays off, you deserve this 🚀

I think that asking people to pay doesn't solve the problem here, if tomorrow thousands of pRs come in, who's going to do the reviews? I think an intra-community exchange of services is more profitable than a marginal payment. (I'm not saying that what Thomas does isn't valuable ^^, it's a matter of strategy)

tomalaforge:
Thanks @svenson95 But the goal is not to be profitable on this project, otherwise I would have never done it this way and I would have created a paid project from the start. I'm just looking for a way to keep this project alive and still find some fun doing it. 
But I'm glad that @alcaidio , @svenson95 and @jdegand are helping me.

tomalaforge:
> > i think that solution with 5€ per review and the lifetime option is really good 👍 so that time and work you put into this project pays off, you deserve this 🚀
> 
> I think that asking people to pay doesn't solve the problem here, if tomorrow thousands of pRs come in, who's going to do the reviews? I think an intra-community exchange of services is more profitable than a marginal payment. (I'm not saying that what Thomas does isn't valuable ^^, it's a matter of strategy)

Do you have something in mind ? I will create a thread. This PR is not suited for this conversation

PR #770, opened 2024-04-13
Answer:1

Arkadiusz123:


apps/angular/projection/src/app/data-access/delete-service.ts
```diff
@@ -0,0 +1,3 @@
+export interface DeleteService {
+  deleteOne(id: number): void;
+}
```
svenson95, line 3: Cool abstraction 👍 but i guess you could also add the other two methods, `addAll()` and `addOne()`.

svenson95:
Good job 🚀

PR #775, opened 2024-04-14
Answer:1 projection

tomalaforge:
solution projection
- with signals and signals queries
- strongly typed template outlet

### DIFF
--- a/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
+++ b/apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
@@ -5,23 +5,22 @@ import {
   randTeacher,
 } from '../../data-access/fake-http.service';
 import { TeacherStore } from '../../data-access/teacher.store';
+import { CardRowDirective } from '../../ui/card/card-row.directive';
 import { CardComponent } from '../../ui/card/card.component';
 import { ListItemComponent } from '../../ui/list-item/list-item.component';
 
 @Component({
   selector: 'app-teacher-card',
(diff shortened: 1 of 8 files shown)


apps/angular/projection/src/app/data-access/student.store.ts
```diff
@@ -1,23 +1,21 @@
...
+  students = signal<Student[]>([]);
 
   addAll(students: Student[]) {
-    this.students.next(students);
+    this.students.set(students);
   }
 
   addOne(student: Student) {
-    this.students.next([...this.students.value, student]);
+    this.students.set([...this.students(), student]);
```
LMFinney, line 15: Why not use `update` here?

AnonyFriday: you can use "update" too to get previous value and add new student, but since u current have the students list, u can also use the set() method

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,40 +1,52 @@
...
     `
-      ::ng-deep .bg-light-red {
+      .bg-light-red {
         background-color: rgba(250, 0, 0, 0.1);
       }
     `,
   ],
   standalone: true,
-  imports: [CardComponent],
+  imports: [ListItemComponent, AsyncPipe, CardRowDirective, CardComponent],
```
abiramcodes, line 32: there is no need for the asyncPipe here, unused import

PR #778, opened 2024-04-16
Answer:1

chekit:

- [x ] Start your PR message with Answer:${challenge_number}

## Warning:

  - $5 per review
  - $25 for lifetime reviews


apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,40 @@
...
+import { City } from '../../model/city.model';
+import { DataCardComponentBase, DataStoreBase } from '../../shared';
+import { CardComponent } from '../../ui/card/card.component';
 
 @Component({
   selector: 'app-city-card',
-  template: 'TODO City',
+  template: `
+    <app-card
+      [list]="data$ | async"
```
GregOnNet, line 13: Cool that you use a stream here.
You should be able to use `ChangeDetectionStrategy.OnPush` here.

`#performance` `#no-blocker`

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,40 @@
...
+      (addNewItem)="onAddNewItem()"
```
GregOnNet, line 14: Instead of `onAddNewItem` you could be more specific about the handler → `addNewCitiy()`.

`#personalPerference` `#no-blocker`

apps/angular/projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,13 +1,40 @@
...
+      (deleteItem)="onDeleteItem($event)"
+      [backgroundColor]="'rgba(250, 10, 250, 0.1)'">
```
GregOnNet, line 16: I think this is better than using `::ng-deep`.

We could also think of allowing to pass a class-name: `backgroundClass="bg-red-200"`

This allows to reuse styles from an existing CSS-Framework like prime flex or tailwind.

`#idea` `#no-blocker`

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
```diff
@@ -1,40 +1,40 @@
...
+export class TeacherCardComponent
+  extends DataCardComponentBase<Teacher>
+  implements OnInit
+{
   ngOnInit(): void {
     this.http.fetchTeachers$.subscribe((t) => this.store.addAll(t));
+  }
 
-    this.store.teachers$.subscribe((t) => (this.teachers = t));
+  override onAddNewItem() {
```
GregOnNet, line 37: OK, now I understand the “generic” naming.

From my experience, this adds another layer of indirection that does not provide a benefit for the developer.
This approach makes absolute sense to me if we are in the area of spinning up components dynamically from a configuration. Then we need to introduce contracts and hooks.

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,48 @@
...
         class="rounded-sm border border-blue-500 bg-blue-300 p-2"
-        (click)="addNewItem()">
+        (click)="this.addNewItem.emit()">
         Add
       </button>
     </div>
   `,
+  styles: `
+    .card__image {
+      width: 200px;
```
GregOnNet, line 34: Why do we need to hard-code the width of the image here, if our goal is to have an abstract card-implementation?

`#api` `#no-blocker`

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,48 @@
...
-      <img
-        *ngIf="type === CardType.TEACHER"
-        src="assets/img/teacher.png"
-        width="200px" />
-      <img
-        *ngIf="type === CardType.STUDENT"
-        src="assets/img/student.webp"
-        width="200px" />
-
+      [bgColor]="backgroundColor()">
```
GregOnNet, line 11: If we allow passing a CSS-class, we do not need the custom `BgColorDirective`.

`#no-blocker` `#less-code-to-maintain`

apps/angular/projection/src/app/ui/card/card.component.ts
```diff
@@ -1,61 +1,48 @@
...
+  `,
   standalone: true,
-  imports: [NgIf, NgFor, ListItemComponent],
+  imports: [NgFor, ListItemComponent, NgStyle, HighlightDirective],
 })
 export class CardComponent {
-  @Input() list: any[] | null = null;
-  @Input() type!: CardType;
-  @Input() customClass = '';
+  list = input<{ id: number; firstName?: string; name?: string }[] | null>(
```
GregOnNet, line 41: This typing limits the data structure we can pass to the card.
How would we act if we would like to use this card for displaying a `Product` having no `firstname`.

Maybe it is possible to make `CardComponent` generic: `CardComponent<T>`

`#api`

GregOnNet (commented):
Hi @chekit,

thank you so much for your PR. 🙏🏻
This is the first review, I do.
So please do not take my comments too serious.

Here are general thoughts about the given goal “. The goal of the challenge is to create a `CardComponent` that can be customized without any modifications.”
- I think the typing of the `list-input` of `CardComponent` would cause an issue, since the typing limits the usage to objects having `id, firstname, name`.
- Other than that, I think the intention was to make the card reusable by using `ngTemplateOutlet`
- Nevertheless, you managed to make the card reusable by abstracting the data-model.
- In the end, I think `ngTemplateOutlet` makes the card more reusable.

Cheers from Leipzig, Germany
Gregor

apps/angular/projection/src/app/component/city-card/city-card.component.ts
chekit: Definitely make sense 👍🏻

apps/angular/projection/src/app/component/city-card/city-card.component.ts
chekit: Yeah, as you noticed later the intention was to move all repetitive code out of components, to make it possible to handle it in one place.

apps/angular/projection/src/app/component/city-card/city-card.component.ts
chekit: I've left it with `style`, but with refactored implementation class could be easily implemented as well :)

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
chekit: The idea is to move all repeated functionality to abstract class (DRY), thus you don't need to copy paste it once you create a new card type component and focus only on the things that differs this component from others.

apps/angular/projection/src/app/ui/card/card.component.ts
chekit: Indeed, put this back to the specific card as it was initially.

apps/angular/projection/src/app/ui/card/card.component.ts
chekit: Yeah, now I see what you mean. It is definitely look like an overhead.

apps/angular/projection/src/app/ui/card/card.component.ts
chekit: You are right. With the ngTemplateOutlet approach it is not need to be too specific for list property.

chekit:
> Hi @chekit,
> 
> thank you so much for your PR. 🙏🏻 This is the first review, I do. So please do not take my comments too serious.
> 
> Here are general thoughts about the given goal “. The goal of the challenge is to create a `CardComponent` that can be customized without any modifications.”
> 
> * I think the typing of the `list-input` of `CardComponent` would cause an issue, since the typing limits the usage to objects having `id, firstname, name`.
> * Other than that, I think the intention was to make the card reusable by using `ngTemplateOutlet`
> * Nevertheless, you managed to make the card reusable by abstracting the data-model.
> * In the end, I think `ngTemplateOutlet` makes the card more reusable.
> 
> **Btw:** I am not sure if I should approve the PR or not. Not sure about the process, here. :-) //cc @tomalaforge
> 
> Cheers from Leipzig, Germany Gregor

Hi @GregOnNet,

Thank you a lot for such a detailed review. It was really nice to practice and recall the concepts which I'm not using everyday. I did some changes and left a comments to your comments.

Once again, much appreciation :)

P.S.
I'm not sure you need to approve PR to not override the original exercise 

Cheers from Amsterdam, The Netherlands. Anton 😉

apps/angular/projection/src/app/component/teacher-card/teacher-card.component.ts
GregOnNet: > The idea is to move all repeated functionality to abstract class (DRY), thus you don't need to copy paste it once you create a new card type component and focus only on the things that differs this component from others.

Yes, I understand the goal. As far as I can tell, this approach only works in very few situations.
Often we tend to abstract too early. Then a business requirement changes, and we start adding conditions to our abstractions in order to make them work for a specific case.

That's why I value SRP (Single Responsibility Principle) over DRY.
In our challenge here, I would ask, what is a card responsible for? Aha, providing a visual boundary and a spot to render 
any content.

GregOnNet:
@chekit Hey, thanks for the feedback. I really appreciate it.

Yes, I learned, that these PRs are tagged in order to mark them as answer. :-)

PR #782, opened 2024-04-17
Answer 48: Avoid losing form data solution

svenson95:
- I used a `canDeactivate` guard to prevent navigation while the form is dirty. 
- Furthermore i've added a `UnloadGuardDirective`, to warn the user before navigate outside the application, like closing the window or navigate to a different page. 

---

There is a problem with angular material, the prepared dialog was shown at the bottom of the page. After adding a theme to the `project.json` (angular.json) `styles` array it works as expected. Maybe we should fix this, to improve the "challenge experience" 😄 

First i've added a `MatDialogModule` inside the `dialog.component`, which was not really necessary - but this lead me to the theme problem. While importing `MatDialogModule`, this warn was logged.
Bildschirmfoto 2024-04-17 um 09 25 07

### DIFF
--- a/apps/forms/form-dialog-alert/project.json
+++ b/apps/forms/form-dialog-alert/project.json
@@ -20,7 +20,10 @@
           "apps/forms/form-dialog-alert/src/favicon.ico",
           "apps/forms/form-dialog-alert/src/assets"
         ],
-        "styles": ["apps/forms/form-dialog-alert/src/styles.scss"],
+        "styles": [
+          "apps/forms/form-dialog-alert/src/styles.scss",
+          "./node_modules/@angular/material/prebuilt-themes/indigo-pink.css"
+        ],
         "scripts": []
       },
       "configurations": {
--- a/apps/forms/form-dialog-alert/src/app/guards/unload-window.directive.ts
+++ b/apps/forms/form-dialog-alert/src/app/guards/unload-window.directive.ts
@@ -1,0 +1,13 @@
+import { Directive, HostListener, inject } from '@angular/core';
+import { Observable } from 'rxjs';
+import { FormComponent } from '../ui/form.component';
+
+@Directive({ selector: 'app-form[appUnloadGuard]', standalone: true })
+export class UnloadGuardDirective {
+  readonly component = inject(FormComponent);
+
+  @HostListener('window:beforeunload')
+  canDeactivate(): Observable<boolean> | boolean {
+    return !this.component.form!.dirty;
+  }
+}
--- a/apps/forms/form-dialog-alert/src/app/pages/join.component.ts
+++ b/apps/forms/form-dialog-alert/src/app/pages/join.component.ts
@@ -1,16 +1,40 @@
-import { ChangeDetectionStrategy, Component } from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
+  inject,
+  viewChild,
+} from '@angular/core';
+import { MatDialog } from '@angular/material/dialog';
+import { Observable, of } from 'rxjs';
+import { PendingChangesGuard } from '../guards/pending-changes.guard';
+import { UnloadGuardDirective } from '../guards/unload-window.directive';
+import { AlertDialogComponent } from '../ui/dialog.component';
 import { FormComponent } from '../ui/form.component';
 
 @Component({
   standalone: true,
-  imports: [FormComponent],
+  imports: [FormComponent, UnloadGuardDirective],
   template: `
     <section class="mx-auto	max-w-screen-sm">
       <div class="rounded-lg bg-white p-8 shadow-lg lg:p-12">
-        <app-form />
+        <app-form appUnloadGuard />
       </div>
     </section>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class JoinComponent {}
+export class JoinComponent implements PendingChangesGuard {
+  private dialog = inject(MatDialog);
+  formComponent = viewChild(FormComponent);
+
+  canDeactivate(): boolean | Observable<boolean> {
+    if (this.formComponent()!.form.dirty) {
+      const dialogRef = this.dialog.open(AlertDialogComponent, {
+        disableClose: true,
+      });
+      return dialogRef.afterClosed();
+    } else {
+      return of(true);
+    }
+  }
+}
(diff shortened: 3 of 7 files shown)


apps/forms/form-dialog-alert/project.json
```diff
@@ -20,7 +20,10 @@
           "apps/forms/form-dialog-alert/src/favicon.ico",
           "apps/forms/form-dialog-alert/src/assets"
         ],
-        "styles": ["apps/forms/form-dialog-alert/src/styles.scss"],
+        "styles": [
+          "apps/forms/form-dialog-alert/src/styles.scss",
+          "./node_modules/@angular/material/prebuilt-themes/indigo-pink.css"
```
svenson95, line 25: this was needed to fix the material dialog styling

apps/forms/form-dialog-alert/src/app/pages/join.component.ts
```diff
@@ -1,16 +1,40 @@
...
 })
-export class JoinComponent {}
+export class JoinComponent implements PendingChangesGuard {
+  private dialog = inject(MatDialog);
+  formComponent = viewChild(FormComponent);
+
+  canDeactivate(): boolean | Observable<boolean> {
+    if (this.formComponent()!.form.dirty) {
+      const dialogRef = this.dialog.open(AlertDialogComponent, {
+        disableClose: true,
```
svenson95, line 33: `disableClose: true` to avoid closing the dialog with escape for example

apps/forms/form-dialog-alert/src/app/pages/join.component.ts
```diff
@@ -1,16 +1,40 @@
...
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class JoinComponent {}
+export class JoinComponent implements ComponentCanDeactivate {
+  private dialog = inject(MatDialog);
+  formComponent = viewChild(FormComponent);
+
+  canDeactivate(): boolean | Observable<boolean> {
+    if (this.formComponent()!.form.dirty) {
```
jdegand, line 31: This gives a `Forbidden non null assertion` eslint error.  You can change it to `this.formComponent()?.form.dirty`.

apps/forms/form-dialog-alert/src/app/guards/unload-window.directive.ts
```diff
@@ -0,0 +1,13 @@
...
+import { Observable } from 'rxjs';
+import { FormComponent } from '../ui/form.component';
+
+@Directive({ selector: 'app-form[appUnloadGuard]', standalone: true })
+export class UnloadGuardDirective {
+  readonly component = inject(FormComponent);
+
+  @HostListener('window:beforeunload')
+  canDeactivate(): Observable<boolean> | boolean {
+    return !this.component.form!.dirty;
```
jdegand, line 11: This gives `Forbidden non null assertion` eslint error.   You can replace it with `!this.component.form?.dirty`.

apps/forms/form-dialog-alert/src/app/pages/join.component.ts
```diff
@@ -1,16 +1,40 @@
...
+        <app-form appUnloadGuard />
       </div>
     </section>
   `,
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class JoinComponent {}
+export class JoinComponent implements ComponentCanDeactivate {
+  private dialog = inject(MatDialog);
+  formComponent = viewChild(FormComponent);
```
jdegand, line 28: If you use `viewChild.required` instead, you won't have to do the extra check ?

apps/forms/form-dialog-alert/project.json
jdegand, line 25: When I served untouched challenge, I didn't have an error in browser console without the theme installed.  Could be necessary to add challenge anyway?

svenson95: Yes the error (warn to be precise) only occurs if you import `MatDialogModule` for example. I looked at the challange again, in the tips section there is a hint with a nicer solution, so i'll refactor this.

> Use the Material CDK Dialog or Overlay - https://material.angular.io/cdk/ (dont forget to add @import '@angular/cdk/overlay-prebuilt.css' in style.sccss)

svenson95: And what do you mean with "Could be necessary to add challenge anyway?"

apps/forms/form-dialog-alert/src/app/ui/dialog.component.ts
```diff
@@ -1,24 +1,34 @@
...
   standalone: true,
+  imports: [MatDialogClose],
   template: `
-    <div role="alert" class="rounded-xl border border-gray-100 bg-white p-5">
-      <h3 class="block text-xl font-medium text-red-600">
-        You have unsaved information!
+    <div
+      role="alertdialog"
+      aria-labelledby="alert-title"
+      aria-describedby="alert-description"
```
svenson95, line 13: The `aria-labelledby` attribute is generally the id of the element used to title the alertdialog.
The `aria-describedby` attribute is generally the ID of the element containing the alert dialog's messaging, usually coming right after the title.

jdegand: All interactive elements must have an accessible name. aria-labelledby can be used to reference another element to define its accessible name, when an element's accessible name needs to use content from elsewhere in the DOM. - [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby).

This leads to believe that you should just use `aria-label` on each button of the modal instead.

apps/forms/form-dialog-alert/project.json
tomalaforge: yes we can add it to the starter file. That's not part of the challenge to find out how to import the theme

tomalaforge:
Great work, but you can rebase your challenge because of all the change you have made 😅

svenson95:
> Great work, but you can rebase your challenge because of all the change you have made 😅

I think i did something wrong, now there are 41 files changed. Do you know how to fix this?

jdegand:
When you added your other PR, you must have changed the file names and now there is a merge conflict.  Maybe, delete the files, merge and then add them back ?

PR #784, opened 2024-04-17
Answer:43

tsironis13:


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,24 @@ const ageToCategory = (age: number): Category => {
...
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
+  age = input(0, {
+    transform: (value: string): number => Number(value),
+  });
```
svenson95, line 35: Simply use `transform: numberAttribute`, imported from `'@angular/core'` 🚀

svenson95 (approved):
👍

apps/angular/signal-input/src/app/user.component.ts
tsironis13: fixed...ty

PR #786, opened 2024-04-17
Answer:46

tsironis13:


apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -17,72 +27,70 @@ import { Component } from '@angular/core';
...
-          <span>28</span>
-        </div>
-
-        <div class="list-item">
-          <span>Birthdate:</span>
-          <span>02.11.1995</span>
-        </div>
+  templateUrl: 'app.component.html',
+  animations: [
+    trigger('listAnimation', [
```
svenson95, line 32: I've used a file `app.animation.ts` for this animation in my solution, so it looks like this `animations: [LIST_ANIMATION]`. To keep the component cleaner, less lines of code.

apps/angular/simple-animations/src/app/component/paragraph.component.html
```diff
@@ -0,0 +1,6 @@
+<div [@movingAnimation]>
```
svenson95, line 1: unnecessary brackets `[ ]` but that's fine

svenson95:
Looks good 👍 good job

PR #787, opened 2024-04-17
Answer:5

Sanjar1304:
Added ngrx state management and testing

### DIFF
--- a/apps/angular/crud/src/app/ui/todo/todo.component.html
+++ b/apps/angular/crud/src/app/ui/todo/todo.component.html
@@ -1,0 +1,20 @@
+
+<div class="flex items-center justify-between gap-5 p-2 hover:bg-slate-200">
+  <div class="w-30vw align-middle">
+    {{todo.title}}
+  </div>
+  <div class="flex gap-3">
+    <button class="bg-blue-500 hover:bg-blue-700 text-white  py-2 px-4 rounded-full hover:shadow-lg
+      {{loaderService.isLoading() ? 'opacity-50 cursor-not-allowed' : ''}}"
+      (click)="updateTodo(todo)"
+      [disabled]="loaderService.isLoading()">
+      Update
+    </button>
+    <button class="bg-pink-500 hover:bg-pink-700 text-white  py-2 px-4 rounded-full hover:shadow-lg
+      {{loaderService.isLoading() ? 'opacity-50 cursor-not-allowed' : ''}}"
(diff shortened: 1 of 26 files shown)


Sanjar1304:
Added ngrx store
Screenshot 2024-04-17 192951
-devtools. Is it ok ??? that's why branch is giving conflict.

apps/angular/crud/src/app/ui/todo/todo.component.html
```diff
@@ -0,0 +1,20 @@
...
+  </div>
+  <div class="flex gap-3">
+    <button class="bg-blue-500 hover:bg-blue-700 text-white  py-2 px-4 rounded-full hover:shadow-lg
+      {{loaderService.isLoading() ? 'opacity-50 cursor-not-allowed' : ''}}"
+      (click)="updateTodo(todo)"
+      [disabled]="loaderService.isLoading()">
+      Update
+    </button>
+    <button class="bg-pink-500 hover:bg-pink-700 text-white  py-2 px-4 rounded-full hover:shadow-lg
+      {{loaderService.isLoading() ? 'opacity-50 cursor-not-allowed' : ''}}"
```
svenson95, line 14: I'd rather refactor that part `{{loaderService.isLoading() ? 'opacity-50 cursor-not-allowed' : ''}}` 
with something like this `[ngClass]="{ 'opacity-50 cursor-not-allowed': loaderService.isLoading() }"`

svenson95:
Looks fine for me 🥇

apps/angular/crud/src/app/ui/todo/todo.component.html
Sanjar1304: Thanks for attention and advice. I did not think of this option. Nice solution instead of using string interpolation.

tomalaforge: ngClass will get deprecated and is not very optimized, 
you should use `[class.opacity-50]="loaderService.isLoading()"` and so on.

tomalaforge:
look fine for me but ngrx is so much boilerplate, look at tanstack query 😅

apps/angular/crud/src/app/ui/todo/todo.component.html
svenson95: or add `[class.is-loading]="loaderService.isLoading()"` and in styles `.is-loading { @apply opacity-50 cursor-not-allowed; }`

PR #798, opened 2024-04-20
Answer:43

Sanjar1304:
added input signals and added transform attribute

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -2,8 +2,9 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
-  OnChanges,
+  computed,
+  input,
+  numberAttribute,
 } from '@angular/core';
 
 type Category = 'Youth' | 'Junior' | 'Open' | 'Senior';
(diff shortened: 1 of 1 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,20 @@ const ageToCategory = (age: number): Category => {
...
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
+  age = input<number, unknown>(0, {
+    transform: numberAttribute,
```
svenson95, line 34: that's clean 👍

svenson95 (approved):
looks-good-to-sxlux9

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +20,20 @@ const ageToCategory = (age: number): Category => {
...
   },
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input<string>();
```
tomalaforge, line 32: you can add a default value to avoid the `${this.lastName() ?? ''}` but otherwise that's great

PR #807, opened 2024-04-23
Answer: 48 (author solution)

alcaidio:
I dont know how to add me in author solution in the challenge page? Do i need to do a separate PR?


svenson95:
@alcaidio thomas will add a label `author answer`, thereby it will be added as the authors solution

apps/forms/avoid-losing-form-data/src/app/ui/dialog.component.ts
```diff
@@ -14,11 +13,13 @@ import { ChangeDetectionStrategy, Component } from '@angular/core';
 
       <div class="mt-4 flex gap-2">
         <button
+          (click)="dialogRef.close(true)"
```
svenson95, line 16: You can simply use the directives `[matDialogClose]="true"` and `[matDialogClose]="false"`, so you don't need the dialogRef injection at all 🚀

apps/forms/avoid-losing-form-data/src/app/ui/form.component.ts
```diff
@@ -63,16 +71,33 @@ import { FormBuilder, ReactiveFormsModule, Validators } from '@angular/forms';
...
+  constructor() {
+    this.form.statusChanges
+      .pipe(
+        map((form) => Object.values(form)),
+        map((values) => values.map((v) => v.trim()).some((v) => v.length)),
+        distinctUntilChanged(),
+        tap((hasValues) => this.hasValues.emit(hasValues)),
+        takeUntilDestroyed(),
+      )
+      .subscribe();
```
tomalaforge, line 94: why don't you listen to form.dirty, form.touched and add a submit flag ? 
it seems easier to understand 😅
but your solution works as well. But you will have to write it in all your form components.

tomalaforge:
Thanks for submitting your solution 🙏

apps/forms/avoid-losing-form-data/src/app/ui/form.component.ts
alcaidio: Because a form can still be marked as dirty even if the user starts typing and manually cleans the form afterward, if I remember correctly?

"But you will have to write it in all your form components." maybe i could create an RxJS operator for that

PR #809, opened 2024-04-23
Answer:32 # refactor(navigation): optimize navigation change detection

sefatanam:
## Checklist:

- Add `OnPush` change detection strategy.
- Remove function calls from the template.
- Use pure pipes instead of function calls, as pure pipes won't be called twice with the same parameter.

## Warning:

### DIFF
--- a/apps/angular/bug-cd/src/app/menus.pipe.ts
+++ b/apps/angular/bug-cd/src/app/menus.pipe.ts
@@ -1,0 +1,18 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'menus',
+  standalone: true,
+})
+export class MenusPipe implements PipeTransform {
+  transform(prop: string) {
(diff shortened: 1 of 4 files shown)


apps/angular/bug-cd/src/app/menus.pipe.ts
```diff
@@ -0,0 +1,18 @@
...
+@Pipe({
+  name: 'menus',
+  standalone: true,
+})
+export class MenusPipe implements PipeTransform {
+  transform(prop: string) {
+    return this.getMenu(prop);
+  }
+
+  getMenu(prop: string) {
```
svenson95, line 12: That's a little bit over-engineered. The `transform` function can return his string array too, no need to create a extra function for it, but that's just my opinion. But I'd define the return type of the `transform` function, which should be `string[]`.

sefatanam: I got your point

PR #811, opened 2024-04-24
Answer:10 Strongly Typed Pipe

LKL2017:


apps/angular/pipe-hard/src/app/util-func.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'utilFunc',
+  standalone: true,
+})
+export class UtilFuncPipe implements PipeTransform {
```
LMFinney, line 7: This is really good as a generic pipe, but it doesn't fit the requirements:
`The goal is to create a specific pipe for this utils file, where you will need to pass the name of the function you want to call and the needed arguments.`

PR #814, opened 2024-04-25
Answer: 46

Sanjar1304:
added angular animations

### DIFF
--- a/apps/angular/simple-animations/src/app/app.component.ts
+++ b/apps/angular/simple-animations/src/app/app.component.ts
@@ -1,8 +1,18 @@
+import {
+  animate,
+  query,
+  stagger,
+  style,
+  transition,
+  trigger,
+} from '@angular/animations';
 import { Component } from '@angular/core';
+import { ItemsInterface } from './items.interface';
+import { ParagraphComponent } from './paragraph.component';
 
 @Component({
   standalone: true,
-  imports: [],
+  imports: [ParagraphComponent],
   selector: 'app-root',
   styles: `
     section {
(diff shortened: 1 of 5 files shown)


apps/angular/simple-animations/src/app/paragraph.component.ts
```diff
@@ -0,0 +1,24 @@
...
+          '0.10s ease-in-out',
+          style({ transform: 'translateX(10%)', opacity: 1 }),
+        ),
+      ]),
+    ]),
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class ParagraphComponent {
+  data = input.required<{ title: string; content: string }>();
```
svenson95, line 23: why you don't use the type `ParagraphItem` here?

apps/angular/simple-animations/src/app/paragraph.component.ts
```diff
@@ -0,0 +1,24 @@
+import { animate, style, transition, trigger } from '@angular/animations';
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
+
+@Component({
+  standalone: true,
+  selector: 'app-paragraph',
+  templateUrl: 'paragraph.component.html',
+  styles: [],
+  animations: [
+    trigger('myAnimation', [
```
svenson95, line 10: You can refactor this animation code to a new `paragraph.animation.ts` file, i like this approach because it reduces the lines of code for this component.

apps/angular/simple-animations/src/app/paragraph.component.html
```diff
@@ -0,0 +1,4 @@
+<div [@myAnimation]="data">
```
svenson95, line 1: Do you need this data reference to get the animation work? maybe just `@myAnimation` works too

apps/angular/simple-animations/src/app/app.component.ts
```diff
@@ -20,69 +30,85 @@ import { Component } from '@angular/core';
...
-            quidem sit placeat possimus tempore dolorum inventore corporis atque
-            quae ad, nobis explicabo delectus.
-          </p>
-        </div>
+        @for (item of items.paragraphItem; track $index) {
+          <div>
+            <app-paragraph
+              [data]="{
+                title: item.title,
+                content: item.content
```
svenson95, line 38: don't belong to the challenge tasks, but you could just pass the `item` object

svenson95 (approved):
good job 🚀

PR #817, opened 2024-04-26
Answer:36

Sanjar1304:
added ngfor in new syntax

### DIFF
--- a/apps/performance/ngfor-optimize/src/app/person-list.component.ts
+++ b/apps/performance/ngfor-optimize/src/app/person-list.component.ts
@@ -1,37 +1,38 @@
-import { Component, EventEmitter, Input, Output } from '@angular/core';
+import { Component, input, output } from '@angular/core';
 
-import { CommonModule } from '@angular/common';
 import { Person } from './person.model';
 
 @Component({
   selector: 'app-person-list',
   standalone: true,
-  imports: [CommonModule],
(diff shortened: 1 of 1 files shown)


apps/performance/ngfor-optimize/src/app/person-list.component.ts
```diff
@@ -1,37 +1,38 @@
...
-          class="rounded-md border bg-blue-500 p-2 text-white"
-          (click)="update.emit(person.email)">
-          UPDATE
-        </button>
-        <button
-          class="rounded-md border bg-red-500 p-2 text-white"
-          (click)="delete.emit(person.email)">
-          DELETE
-        </button>
+    @for (person of persons(); track $index) {
```
tomalaforge, line 10: 👍 it's easier with the new control flow since the trackby is mandatory

However it's better to target a unique property of the person instead of the $index. $index works well if the list if not updated

Sanjar1304: Thanks for recommendation. I have heard that $index property can be used when there is no unique property of the object. In such case I can use $index. In other conditions it is better to use unique properties of the iterating object, such as id or something else. New syntax of structural directives is easier to manage template flow when the condition is required to insert them, without using ngcontainer or ngtemplate.

PR #841, opened 2024-05-03
Challenge 50: bug with conditional effects

tomalaforge:


tomalaforge:
@svenson95 @jdegand Here is a new challenge on signal, I will try to create more of them but it's a very easy one to start.
So you have feedback on the doc page or the challenge itself 🙏

jdegand:
@tomalaforge  Like the challenge a lot.  There are multiple solutions you can employ to get the alert to display after every click.  This is a challenge that is easy in retrospect and requires you to dig deeper and think how `effect` works.  

Maybe, add template literal to the alert ?  Change `alert('One checkbox is check');` to `{age} checkbox was checked` etc.  
Either way, I think the alert message needs to changed but that's about it.

apps/angular/bug-effect-signal/README.md
```diff
@@ -0,0 +1,13 @@
+# Bug in Effect ?
+
+> author: thomas-laforge
+
+### Run Application
+
+```bash
+npx nx serve angular-bug-effect-signal
```
svenson95, line 8: Maybe `angular-signal-effect-bug` would fit better for the naming, to have `signal` as a prefix for other signal challenges as well, like the pipe challenges, which start with `pipe` as a prefix.

And in the title `Effect in Bug ?` i'd remove the space between `Bug` and `?`. Or simply `Effect Bug` would be nicer imo. I'd avoid question marks in the title. Because until now there is no title with a question mark, which is not really a problem, but it looks cleaner for me.

PR #843, opened 2024-05-03
Answer:43 signal input using transform, alias, and computed

LMFinney:

## Warning:

### DIFF
--- a/apps/angular/signal-input/src/app/user.component.ts
+++ b/apps/angular/signal-input/src/app/user.component.ts
@@ -2,7 +2,7 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
+  input,
   OnChanges,
 } from '@angular/core';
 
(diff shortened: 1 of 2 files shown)


apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -19,23 +19,20 @@ const ageToCategory = (age: number): Category => {
...
-export class UserComponent implements OnChanges {
-  @Input({ required: true }) name!: string;
-  @Input() lastName?: string;
-  @Input() age?: string;
+export class UserComponent {
+  name = input.required<string>();
+  lastName = input('');
+  category = input.required({
+    alias: 'age',
+    transform: (age: string): Category => ageToCategory(Number(age) ?? 0),
```
tomalaforge, line 34: it looks smart but I'm not a big fan of it. It's harder to read, which means harder to maintain. If someone or your future you come bacl to the component, it will take you a minute or 2 to understand what the compoennt is expecting and why you did it like that, you should prefer an age input and then a category compute. Easy to read, to maintain and to extend

LMFinney: I'm sympathetic to the argument that this approach is too clever. I pushed another version.

apps/angular/signal-input/src/app/user.component.ts
```diff
@@ -29,10 +30,8 @@ const ageToCategory = (age: number): Category => {
 export class UserComponent {
   name = input.required<string>();
   lastName = input('');
-  category = input.required({
-    alias: 'age',
-    transform: (age: string): Category => ageToCategory(Number(age) ?? 0),
-  });
+  age = input.required({ transform: numberAttribute });
 
+  category = computed(() => ageToCategory(this.age()));
```
tomalaforge, line 35: Don't you think this is way easier to read ?

LMFinney: You're right, but I was having fun playing with the alias and transform for the first time :)

PR #858, opened 2024-05-07
Answer:50

jdegand:
This solution exploits how JavaScript works and how it evaluates variables. JavaScript looks inside code blocks for the variables it needs, and if it doesn't find that variable, it will go to the next code block until it finds the variable. When learning JavaScript, you should have run into this when learning about `hoisting`. 

So to get the effect to run with the latest value of each signal, you just have to use each signal above the conditional.  The common convention is to create a variable to hold the signal value.

Some notes: 
- `effect` is async.
- Effects keep track of their dependencies dynamically, and only track signals that were read in the most recent execution.
- You can't tell Angular what dependencies to track, like React. React's `useEffect` has a dependency array that you can add variables to and cause re-renders when one of those variables changes. 
- When an OnPush component uses a signal's value in its template, Angular will track the signal as a dependency of that component. When that signal is updated, Angular automatically marks the component to ensure it gets updated the next time change detection runs.
-  You shouldn't convert `effect` to `computed` because you are not returning a signal.

### DIFF
--- a/apps/angular/bug-effect-signal/src/app/app.component.ts
+++ b/apps/angular/bug-effect-signal/src/app/app.component.ts
@@ -28,6 +28,10 @@ export class AppComponent {
 
   constructor() {
     effect(() => {
+      console.log(this.name());
+      console.log(this.age());
+      console.log(this.address());
+
       if (this.name() || this.age() || this.address()) {
         alert('Checkbox was checked');
       }


apps/angular/bug-effect-signal/src/app/app.component.ts
```diff
@@ -28,6 +28,10 @@ export class AppComponent {
 
   constructor() {
     effect(() => {
+      console.log(this.name());
+      console.log(this.age());
+      console.log(this.address());
```
tomalaforge, line 33: This is a way to make the effect run, but you don't want to console log in production application.

jdegand: No, this isn't a real solution, but something I stumbled upon while working on the challenge.  And it shows the basic logic of what's happening.  

I actually think the best solution is to duplicate code and have multiple effects.

```
    effect(() => {
      if (this.name()) {
        alert('One checkbox is check');
      }
    });

  // etc
```
When you uncheck the box, the effect does not run again when you have multiple effects.

jdegand:
> What do you mean by effect is async ? Why not using computed here ?

Effect is correct for this use case, as it is meant for side effects and logs.  Imagine we are debugging and making sure the checkbox model input implementation works correctly.
`Effects always execute asynchronously, during the change detection process` per the [docs](https://angular.io/guide/signals#effects).

apps/angular/bug-effect-signal/src/app/app.component.ts
tomalaforge: yes I think that's the best solution. Easy to read, easy to maintain, easy to extend

jdegand: No, it is because of the short-circuiting that's why multiple effects are better for this challenge.  

Although `effect` can track multiple dependencies, it is really wonky (this challenge shows a edge case) and I think it is better to be certain that your effect will run when your signal changes.

tomalaforge: I agree with @jdegand. my solution will be to have one effect per signal, if there are independent of each other. We should always think of simplicity rather than cleverness, even if we need to repeat some part.

phhien203: Yep. The problem is due to short-circuiting. This Signal Graph clearly visualizes this buggy behavior.

https://github.com/user-attachments/assets/0677b0b2-8bdd-40f0-859e-00ea0beb972b

PR #864, opened 2024-05-08
Answer:50 bug effect signal

LMFinney:

## Warning:

### DIFF
--- a/apps/angular/bug-effect-signal/src/app/app.component.ts
+++ b/apps/angular/bug-effect-signal/src/app/app.component.ts
@@ -27,8 +27,30 @@ export class AppComponent {
   address = model(false);
 
   constructor() {
+    // Assuming I correctly understand that the goal is to show the alert
+    // whenever any of them is checked (independent of whether any others are
+    // selected), and that we shouldn't show the alert when any of them
+    // unchecked, then each signal needs to be considered separately.
+    //
+    // The problem with the previous implementation is that `||` is
+    // short-circuiting; if the first signal is true, the second and third
+    // signals are never checked, so we don't get the desired effect on
+    // subsequent checkboxes.
+    //
+    // If JavaScript had the '|' operator like Java, then we could use that
+    // instead.
     effect(() => {
-      if (this.name() || this.age() || this.address()) {
+      if (this.name()) {
+        alert('Checkbox was checked');
+      }
+    });
+    effect(() => {
+      if (this.age()) {
+        alert('Checkbox was checked');
+      }
+    });
+    effect(() => {
+      if (this.address()) {
         alert('Checkbox was checked');
       }
     });


apps/angular/bug-effect-signal/src/app/app.component.ts
```diff
@@ -27,8 +27,30 @@ export class AppComponent {
...
   constructor() {
+    // Assuming I correctly understand that the goal is to show the alert
+    // whenever any of them is checked (independent of whether any others are
+    // selected), and that we shouldn't show the alert when any of them
+    // unchecked, then each signal needs to be considered separately.
+    //
+    // The problem with the previous implementation is that `||` is
+    // short-circuiting; if the first signal is true, the second and third
+    // signals are never checked, so we don't get the desired effect on
+    // subsequent checkboxes.
```
svenson95, line 38: That's what i struggled with too. In my opinion the goal of this challenge is to understand why this effect works, how it works. And from a user perspective, the alert should be displayed when a checkbox is about to become checked. But that doesn't matter because this challenge is about this particular bug.

LMFinney: Your solution definitely fixed the effects being short-circuited. But then it also had the alert showing in some cases when a checkbox is unchecked. I'm not sure if that's acceptable or a different bug.

apps/angular/bug-effect-signal/src/app/app.component.ts
```diff
@@ -27,8 +27,30 @@ export class AppComponent {
...
+    //
+    // If JavaScript had the '|' operator like Java, then we could use that
+    // instead.
     effect(() => {
-      if (this.name() || this.age() || this.address()) {
+      if (this.name()) {
```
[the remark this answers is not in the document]
jdegand: @LMFinney has a point with the short-circuiting.  All it takes is one `true` for the alert to run.  @svenson95 your solution is valid and correct, but because of the short-circuiting here -> multiple effects is the correct answer for the way the challenge is presently constructed.

apps/angular/bug-effect-signal/src/app/app.component.ts
jdegand: I think this was unintended.   

Yes, the variable solution will display an alert anytime there is one true value.  All the values are tracked.

tomalaforge:
ohhhh 😅 you are going too far 😂. I didn't think at all this use cases at all. 

My goal was just to understand why the effect is not called when the first signal is true. Maybe the message in the alert should be different.

LMFinney:
> ohhhh 😅 you are going too far 😂. I didn't think at all this use cases at all.
> 
> My goal was just to understand why the effect is not called when the first signal is true. Maybe the message in the alert should be different.

No problem :)

Since this was released just yesterday, I'll count myself as just a beta user ;)

Perhaps a sequence description like what I used in https://github.com/tomalaforge/angular-challenges/pull/864#discussion_r1594459110 would be good to show what the acceptance criteria should be.

I'm curious if there's a way to match the behavior of my implementation with a single effect. I don't think there is.

tomalaforge:
> > ohhhh 😅 you are going too far 😂. I didn't think at all this use cases at all.
> > My goal was just to understand why the effect is not called when the first signal is true. Maybe the message in the alert should be different.
> 
> No problem :)
> 
> Since this was released just yesterday, I'll count myself as just a beta user ;)
> 
> Perhaps a sequence description like what I used in [#864 (comment)](https://github.com/tomalaforge/angular-challenges/pull/864#discussion_r1594459110) would be good to show what the acceptance criteria should be.
> 
> I'm curious if there's a way to match the behavior of my implementation with a single effect. I don't think there is.

Or to make thing simpler, just show the alert if at least one checkbox is checked. But the alert should be displayed at each click if at least one checkbox is clicked. So we can keep one effect, or multiple effect or a computed. 

My goal is not about the use case but about the undesirable side-effect

LMFinney:
> Or to make thing simpler, just show the alert if at least one checkbox is checked. But the alert should be displayed at each click if at least one checkbox is clicked. So we can keep one effect, or multiple effect or a computed.
> 
> My goal is not about the use case but about the undesirable side-effect

For that simple description, show the alert on both checking or unchecking if at least one of the three is checked? Yeah, that would work :)

svenson95:
Maybe it's easier to understand, if we use a better example. Imagine you try to buy a laptop, and each checkbox is a extra which can be added - more drive space, ram, better gpu. If one checkbox gets checked, the alert should display "Price increased". If you deselect all checkboxes, no alert should be displayed, since all extras are unselected.

LMFinney:
> Maybe it's easier to understand, if we use a better example. Imagine you try to buy a laptop, and each checkbox is a extra which can be added - more drive space, ram, better gpu. If one checkbox gets checked, the alert should display "Price increased". If you deselect all checkboxes, no alert should be displayed, since all extras are unselected.

I like that. A good motivated example like that would really help.

PR #866, opened 2024-05-08
Answer:9 function pipe

LMFinney:

## Warning:

### DIFF
--- a/apps/angular/pipe-intermediate/src/app/wrap-fn.pipe.ts
+++ b/apps/angular/pipe-intermediate/src/app/wrap-fn.pipe.ts
@@ -1,47 +1,11 @@
 import { Pipe, PipeTransform } from '@angular/core';
 
 @Pipe({ standalone: true, name: 'wrapFn' })
-// I tried WrapFnPipe<Result> here instead of defining the Result type param
-// over and over, and I was surprised that the type safety was lost!
 export class WrapFnPipe implements PipeTransform {
-  transform<Result, ARG>(fn: (arg: ARG) => Result, arg: ARG): Result;
-  transform<Result, ARG1, ARG2>(
-    fn: (arg1: ARG1, arg2: ARG2) => Result,
-    arg1: ARG1,
-    arg2: ARG2,
-  ): Result;
-  transform<Result, ARG1, ARG2, ARG3>(
-    fn: (arg1: ARG1, arg2: ARG2, arg3: ARG3) => Result,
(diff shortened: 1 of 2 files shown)


apps/angular/pipe-intermediate/src/app/wrap-fn.pipe.ts
```diff
@@ -0,0 +1,47 @@
...
+      ...args: unknown[]
+    ) => Result,
+    arg1: ARG1,
+    arg2: ARG2,
+    arg3: ARG3,
+    arg4: ARG4,
+    ...args: unknown[]
+  ): Result;
+
+  transform<Result>(
```
LMFinney, line 41: Getting this basic function was the easy part. When I saw that type safety was required, I hoped there was a way to use https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype to handle this, but nope. Ugh.

tomalaforge: ouch, lots of function override. 

You can do shorter 😅

LMFinney: I saw your solution had one less override (resorting to `any[]` earlier). Is that what you mean? Or another idea?

tomalaforge:
```
export class WrapFnPipe implements PipeTransform {
  transform<R, F extends (...args: any[]) => R>(
    func: F,
    ...args: Parameters<F>
  ): R {
    return func(...args);
  }
}
```

LMFinney: Yeah, but that loses all type-safety, right? I think I had that at first, but thought we were supposed to impose type safety.

LMFinney: I tried it, and you're right. There were two parts I didn't catch when I just read your comment:
1. Using `Parameter<F>` is the key - it infers the args, and I missed that.
2. The args have to be `any[]` instead of `unknown[]`. Using `unknown` causes a compilation error. I wouldn't have predicted that.

Thanks

LMFinney: Actually, `never[]` also works, and linting doesn't mind that.

PR #868, opened 2024-05-09
feat(c50): update challenge example

svenson95:
Here is my feedback and additions for this new challenge.


svenson95:
@tomalaforge as mentioned [here](https://github.com/tomalaforge/angular-challenges/pull/841#discussion_r1592451565), we should think about rename this challenge naming. If you want me to do this, just let me know.

apps/angular/bug-effect-signal/src/app/app.component.ts
```diff
@@ -11,25 +11,42 @@ import { FormsModule } from '@angular/forms';
...
-  age = model(false);
-  address = model(false);
+  drive = model(false);
+  ram = model(false);
+  gpu = model(false);
 
   constructor() {
+    /* 
+      Explain for your junior team mate why this bug occurs ...
+    */
```
tomalaforge, line 46: This comment is not very useful to the challenge.

docs/src/content/docs/challenges/angular/50-bug-effect-signal.md
```diff
@@ -13,10 +14,27 @@ sidebar:
...
+## Acceptance Criteria
+
+To ensure this feature works properly, try this out to reproduce the bug after solving the challenge, to check if the bug is gone.
+
+- Check box 1 (Alert should be shown)
+- Check box 2 (Alert should be shown)
+- Uncheck box 1
+- Check box 3 (Alert should be shown)
+- Uncheck box 2
+- Uncheck box 3
```
tomalaforge, line 36: If we do it this way. the only way to solve the challenge is by having multiple effects. Otherwise even if we uncheck one checkbox, the alert will be shown.

tomalaforge:
Nice use case and if you want to change the title: what title will you use?

apps/angular/bug-effect-signal/src/app/app.component.ts
svenson95: To explain why this error occurs can be helpful to really understand it. But if you don't like it i can remove it.

docs/src/content/docs/challenges/angular/50-bug-effect-signal.md
svenson95: That would be okay for you? if so, we have to update our solutions

svenson95:
> Nice use case and if you want to change the title: what title will you use?

For the title i prefer just "Effect bug", or "Bug in Effect" without the question mark.
The filename should be `signal-effect-bug` or `signal-bug-in-effect` respectively.

docs/src/content/docs/challenges/angular/50-bug-effect-signal.md
tomalaforge: It's just that the challenge was just about the issue on the effect. But this doesn't really matter. We can keep it this way. I just don't know how to do it simply with one effect

tomalaforge:
I updated the project tree, so when this PR is done, I will move the app folder as well

docs/src/content/docs/challenges/angular/50-bug-effect-signal.md
svenson95: Yeah you're right. I suggest to change line 40 to the following. So we can find different ways to solve this, maybe with one effect, i'll try this.

> Try to implement this feature with a `computed` signal.

PR #878, opened 2024-05-10
Answer:16 di

LMFinney:

## Warning:


apps/angular/di/src/app/product-row.component.ts
```diff
@@ -0,0 +1,34 @@
...
+  `,
+  imports: [AsyncPipe, CurrencyPipe],
+  providers: [CurrencyService],
+})
+export class ProductRowComponent {
+  product = input.required<Product>();
+  currencyService = inject(CurrencyService);
+
+  constructor() {
+    effect(
```
LMFinney, line 24: I tried changing `CurrencyService` to be signal-based, but it didn't work with this effect. It worked fine if I used the old `@Input` approach, but the timing was wrong with setting the value on the signal from here. I don't know why.

tomalaforge: yes it's a bit tricky. 

You need to understand the order of the event and what is a pure pipe. 
Pure pipe means, the pipe will memoize the value and not recompute a new value if not inputs changed. 
effect happens after the first change detection cycle, so the service value is the default one (Euro), the effect happens, but the input of the pipe doesn't change, only the internal signal change. 
Three solutions: 
- return a computed signal from the pipe: this way when `this.currencyService.symbol()` change, the signal inside the pipe will get recompute
However the way to write the value inside the template is bizarre: `(product().priceA | currency)()`
- transform your pipe to impure
- Since you created a component, create a computed signal that return the symbol directly

I hope my explanation is clearer for you .

tomalaforge: But this is a super good challenge idea 🔥

LMFinney: This makes sense. I was just surprised that the change detection cycle order is different for an `@Input` and a signal-based input.

I didn't try the impure pipe (yuck), but I got the other ideas to work.

Thanks

PR #904, opened 2024-05-13
challenge 53 big signal performance

tomalaforge:


jdegand:
```
For this challenge, you can imagine a large-scale application where you use a service to save and retrieve your user state at any time within the application.

The problem is that updating a single user property updates the entire application.

I added the `CDFlashingDirective` to visualize when one component is rendering.

## Statement

With signals, you can now be more fine-grained in what the UI is rendering. The goal of this challenge is to understand why everything is re-rendering and to refactor the application to be more performant.
```

I corrected the typos.  

Constraints?  Without any, there should be some interesting solutions.   It might be possible to solve this challenge by breaking the best practice of always returning a new object inside `update`.  If you mutate the object, change detection should not run.

jdegand:
This challenge could possibly be enhanced by incorporating a type safety requirement with new `@let` syntax.  Here's an [article](https://netbasal.com/exploring-angulars-new-let-syntax-enhancing-template-variable-declarations-40487b022b44) about it.

tomalaforge:
> This challenge could possibly be enhanced by incorporating a type safety requirement with new `@let` syntax. Here's an [article](https://netbasal.com/exploring-angulars-new-let-syntax-enhancing-template-variable-declarations-40487b022b44) about it.

What do you mean ?

jdegand:
With an object inside a signal, you lose type safety in the template when call  `signal(). property`.  `@let` allows you to type nested properties by declaring the nested property as a variable.  

However, in this challenge, I guess most solutions will change what is returned in the template so I guess it doesn't make sense to add it here.   This was first challenge that came to mind where something like could be added versus adding a separate challenge for it.

tomalaforge:
> With an object inside a signal, you lose type safety in the template when call `signal().property`. `@let` allows you to type nested properties by declaring the nested property as a variable.
> 
> However, in this challenge, I guess most solutions will change what is returned in the template so I guess it doesn't make sense to add it here. This was first challenge that came to mind where something like could be added versus adding a separate challenge for it.

You don't loose type safety 🤔

The only thing that is not working with signal currently is type narrowing. 
like 
```
@if(user()) {
 // user is still of type User and undefined
}
```

jdegand:
You are right.  I was thinking of type narrowing.

PR #907, opened 2024-05-13
Answer:47 unions

LMFinney:
Comparing mine to yours (now that I'm done), mine is simpler, which makes me think I missed some of what you intended with this. Since the keys and values are the same, the "Author's solution" seems to add more concepts than needed for `Difficulty`.

Am I missing something?

## Warning:


tomalaforge:
yes indeed your solution is simpler. 

On my case, i just wanted to be closer to the first implementation with enums. And your solution works because the union type and the label you display are identical.

PR #912, opened 2024-05-14
Answer:2 effect vs selector

LMFinney:
The description said to use only one action from the component, but I used the effect lifecycle instead, so the component dispatches nothing.
Also, I used `createFeature` to simplify the selectors.

## Warning:


apps/ngrx/2-effect-vs-selector/src/app/store/status/status.selectors.ts
```diff
@@ -1,17 +1,49 @@
...
+            ...status,
+            { name: activity.type, teachers: [activity.teacher] },
+          ];
+        } else {
+          status[index].teachers.push(activity.teacher);
+          return status;
+        }
+      }, []);
+    }
+    return [];
```
tomalaforge, line 31: 👍

tomalaforge:
Good job. The goal of this challenge is to show that selectors are very powerful and the used to derive the state. Like computed in signals. We shouldn't store derived piece of state.

PR #913, opened 2024-05-14
Answer:7 Power of effects

LMFinney:
I'm not sure if I achieved step 4 - I'm not sure what you intended.
We could have injected the PushService directly wherever the injection token was used, right?

## Warning:


jdegand:
You don't have to worry about lazy loading because you used the lifecycle methods of component store.  
In the initial app, all the lists are constantly growing.

The injection token approach makes more sense if you don't (or can't) convert all components to use Component Stores.

apps/ngrx/7-power-of-effect/src/app/app.effects.ts
```diff
@@ -0,0 +1,57 @@
...
+  displayAlerts$ = createEffect(
+    () =>
+      this.actions$.pipe(
+        ofType(
+          schoolActions.addOneSchool,
+          studentActions.addOneStudent,
+          teacherActions.addOneTeacher,
+        ),
+        map((action) => {
+          switch (action.type) {
```
tomalaforge, line 28: I'm not a big fan of switch statement in effect. You should create separate effects for each use cases

apps/ngrx/7-power-of-effect/src/app/push-token.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Push } from '@angular-challenges/power-of-effect/model';
+import { inject, InjectionToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+export const PUSH_TOKEN = new InjectionToken<Observable<Push | undefined>>(
+  'PUSH_TOKEN',
+  {
+    factory: () => inject(PushService).notification$,
+  },
+);
```
tomalaforge, line 11: That's the tricky part and you did it well. Nice job.

apps/ngrx/7-power-of-effect/src/app/school/school.store.ts
```diff
@@ -45,4 +47,20 @@ export class SchoolStore
   ngrxOnStoreInit() {
     this.loadSchools();
   }
+
+  ngrxOnStateInit() {
+    this.effect(() =>
+      inject(PUSH_TOKEN).pipe(
```
tomalaforge, line 53: That's the second point I want to highly with this challenge, it's that we can use any observable inside an effect and not only the Ngrx action observbale. Effect is just a subscription. 

Nice work

tomalaforge:
Very nice work and I think you are the first one to complete the challenge. I was going to suppress it 😅

apps/ngrx/7-power-of-effect/src/app/app.effects.ts
LMFinney: I just pushed two commits:

1. One to split the effect into four effects (1 for each of the three notification types, and one to do the actual alert).
2. Converted to functional effects (just for practice).

apps/ngrx/7-power-of-effect/src/app/app.effects.ts
```diff
@@ -24,10 +20,12 @@ export class AppEffects {
       }),
       map((message) => appApiActions.alert({ message })),
     ),
-  );
+  { functional: true },
+);
```
tomalaforge, line 24: 👍

PR #929, opened 2024-05-17
Answer:32 

Sanjar1304:
fixed cd-bug with for track and signals

### DIFF
--- a/apps/angular/bug-cd/src/app/main-navigation.component.ts
+++ b/apps/angular/bug-cd/src/app/main-navigation.component.ts
@@ -1,5 +1,11 @@
-import { AsyncPipe, NgFor, NgIf } from '@angular/common';
-import { Component, Input, inject } from '@angular/core';
+import {
+  ChangeDetectionStrategy,
+  Component,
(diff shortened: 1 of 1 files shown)


tomalaforge:
Nice, but did you understand what was the issue ?

Sanjar1304:
Without the track in for loop, menu will be created and removed from the start till the end of iterating source and DOM UI will be updated in amount of source. That will affect to the browsers rendering speed. In order to optimize this condition, track function is recommended to use in for loop. Because track funtion only gets unique data of source iterating object and only update the exact changed item of the unique data of source iterating object. It gives us performance optimization by speeding up  DOM UI rendering in browser.  And I also converted data coming from backend to signal. And returning data by checking if data's path is foo or bar, otherwise returning empty string. Signals share changes with each other under the hood effectively. That's why I used them.

tomalaforge:
Indeed trackby will not update the DOM is the updated list have the same element. But this behavior will only hide the real issue in the code. 
Why is the list updated? You should answer this question to understand the real issue

PR #937, opened 2024-05-20
Answer:48

tsironis13:
- Convert dialog component to a dynamic form where each form group controls can be projected giving the flexibility to consumers to provide the desired form group.
- Implement browser native confirm dialog functionality when there are pending form changes.
- It would be really appreciated if someone can come up with a solution to hide/close the material dialog when the page tries to reload using the browser native button before the confirm popup is displayed. Currently, both material dialog and browser native popup are shown at the same time.


jdegand:
As to your last point, here's a [Stack Overflow](https://stackoverflow.com/questions/56325272/detect-browser-refresh-in-an-angular-project) about different ways to detect a browser refresh.  There are a ton of router events and you would probably have to always listen for `NavigationStart` (etc) and then try to call the `dialogRef.close` method.

PR #939, opened 2024-05-21
Answer:51 function-call-effect

jdegand:
This probably isn't correct but it seems to work anyway.

### DIFF
--- a/apps/signal/51-function-call-effect/src/app/action.component.ts
+++ b/apps/signal/51-function-call-effect/src/app/action.component.ts
@@ -4,6 +4,7 @@ import {
   effect,
   inject,
   signal,
+  untracked,
 } from '@angular/core';
 import { FormsModule } from '@angular/forms';
 import { UserService } from './user.service';
@@ -39,7 +40,12 @@ export class ActionsComponent {
 
   constructor() {
     effect(() => {
-      this.userService.log(this.action() ?? 'No action selected');
+      const action = this.action();
+      untracked(() => {
+        this.userService.log(action ?? 'No action selected');
+      });
     });
   }
 }
+
+// https://blog.angulartraining.com/whats-the-untracked-function-angular-signals-c08d28b92d57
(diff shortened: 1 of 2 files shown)


apps/signal/51-function-call-effect/src/app/user.service.ts
```diff
@@ -1,10 +1,18 @@
-import { Injectable, signal } from '@angular/core';
+import { Injectable, signal, untracked } from '@angular/core';
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
-  name = signal('Thomas');
+  name = untracked(signal('Thomas'));
```
tomalaforge, line 5: no, don't use it here. 
If you need to listen to name somewhere else in your application, you will not be able to.
If you want a basic property, just don't use a signal. 😅

apps/signal/51-function-call-effect/src/app/user.service.ts
```diff
@@ -1,10 +1,15 @@
...
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
   name = signal('Thomas');
 
+  // name is considered a dependency by Angular
+
   log(message: string) {
-    console.log(`${this.name()}: ${message}`);
+    console.log(untracked(() => this.name()) + ':' + message);
```
tomalaforge, line 10: I need to update the documentation. Let's say, userService is a third party library, we don't have access to it.

apps/signal/51-function-call-effect/src/app/action.component.ts
```diff
@@ -39,7 +40,12 @@ export class ActionsComponent {
 
   constructor() {
     effect(() => {
-      this.userService.log(this.action() ?? 'No action selected');
+      const action = this.action();
+      untracked(() => {
+        this.userService.log(action ?? 'No action selected');
+      });
```
tomalaforge, line 46: Yes exactly, this is the right solution 💪

PR #942, opened 2024-05-21
Answer:50 bug in effect

tomalaforge:


SidV2 (approved):

elminino21:
sorry, to bother, but this one has me puzzled. I have not been able figure out why this answer works.  my fist thought was the effect doesnt changes because it doesnt have a different output.  However, when I tried to check the values were being updated correctly. 
` effect(() => {
      console.log(this.drive() + " "  +  this.ram()+ " "  +  this.gpu());
      if (this.drive() || this.ram() || this.gpu()) {
        alert('Price increased!');
      }
    });`
for some reason adding the console.log() allows the alert to display as expected.  Any thoughts. 
Thank you for you time in advance :)

PR #946, opened 2024-05-21
Answer:51 signal bug

LMFinney:
Took a break from puzzling over View Transitions to knock this one out :)

## Warning:

### DIFF
--- a/apps/signal/51-function-call-effect/src/app/action.component.ts
+++ b/apps/signal/51-function-call-effect/src/app/action.component.ts
@@ -40,8 +40,8 @@ export class ActionsComponent {
 
   constructor() {
     effect(() => {
-      const name = untracked(() => this.userService.name());
-      console.log(`${name}: ${this.action() ?? 'No action selected'}`);
+      const action = this.action();
+      untracked(() => this.userService.log(action ?? 'No action selected'));
     });
   }
 }
--- a/apps/signal/51-function-call-effect/src/app/action.component.ts
+++ b/apps/signal/51-function-call-effect/src/app/action.component.ts
@@ -4,6 +4,7 @@ import {
   effect,
   inject,
   signal,
+  untracked,
 } from '@angular/core';
 import { FormsModule } from '@angular/forms';
 import { UserService } from './user.service';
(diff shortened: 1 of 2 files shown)


apps/signal/51-function-call-effect/src/app/user.service.ts
```diff
@@ -1,10 +1,10 @@
-import { Injectable, signal } from '@angular/core';
+import { Injectable, signal, untracked } from '@angular/core';
 
 @Injectable({ providedIn: 'root' })
 export class UserService {
   name = signal('Thomas');
 
   log(message: string) {
-    console.log(`${this.name()}: ${message}`);
+    console.log(`${untracked(() => this.name())}: ${message}`);
```
tomalaforge, line 8: I'm not sure, i'll put the untrack here. It might make sense to track name for some other use case. 

I would use the untrack inside the effect since it's where we don't want to track what's inside the function. And for instance, the log can be a library you don't own

LMFinney: My initial commit was the minimal change to fix the bug.

Based on this feedback, I pushed another commit to move the fix to the component.

apps/signal/51-function-call-effect/src/app/action.component.ts
```diff
@@ -39,7 +40,8 @@ export class ActionsComponent {
 
   constructor() {
     effect(() => {
-      this.userService.log(this.action() ?? 'No action selected');
+      const name = untracked(() => this.userService.name());
+      console.log(`${name}: ${this.action() ?? 'No action selected'}`);
```
tomalaforge, line 44: that's still not good. I think I need to change the challenge. You cannot access name and you still need to call the log function. 😅

LMFinney: I thought about it a bit more and had another idea.

Third time's a charm?

apps/signal/51-function-call-effect/src/app/action.component.ts
```diff
@@ -39,7 +40,8 @@ export class ActionsComponent {
 
   constructor() {
     effect(() => {
-      this.userService.log(this.action() ?? 'No action selected');
+      const action = this.action();
+      untracked(() => this.userService.log(action ?? 'No action selected'));
```
tomalaforge, line 44: Yes that's the right solution 👍

PR #987, opened 2024-06-02
feat: add codespace docs

jdegand:
Added some notes about how someone can use GitHub codespaces to run the project and tackle challenges or make a PR.  

I looked into adding  a `.devcontainer/devcontainer.json` file to upgrade the features and node version of the codespace ( initial codespace starts with node 20 and some other basic features already installed) .  

[GitHub link](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers) for basic devcontainer info if you are unfamiliar.  

The codespace installation fails during the automatically run installation process. User needs to run  `npm ci` to correctly install.  

This is why I added a note about how the initial install from Codespaces will fail.  This may be preventable with a `.devcontainer.json` file  and a `Dockerfile` may be necessary to have `npm ci` run automatically.  Still looking more into this.  

I suggest you try out a quick codespace experience and if you have problems, I could add more directions to the readme. 

The codespace doesn't have immediate access to user's clipboard.  Self-explanatory.  

When you serve any project, a box will come up and user can click that to view the app running.  It uses port forwarding.  I didn't add a comment for this.


jdegand:
I think the installation of dependencies only fails if a user rejects the recommended `VS Code` plugins on its prompt.   I'll update the directions to tell challengers to accept the plugins.

jdegand:
I changed the directions.  Apparently, the issues I had installing the codespace were temporary glitches.  It just installed perfectly with and without installing the optional VS Code plugins.  

It would be best if you added a screenshot (as my username would be visible) for the modal with the codespace options (pausing and delete last).  
Screenshot 2024-06-11 11 10 49 PM
  
I added a link for a Github Skills tutorial and I think this is sufficient.

tomalaforge (approved):

PR #993, opened 2024-06-10
docs(ru): 52-lazy-load-component-translation

stillst:
Challenge translation


docs/src/content/docs/ru/challenges/angular/52-lazy-load-component.md
```diff
@@ -0,0 +1,49 @@
+---
+title: 🟢 Ленивая загрузка компонента
+description: Испытание 52 посвящено изучению ленивой загрузки компонентов в Angular.
+author: lance-finney
+contributors:
+  - LMFinney
```
tomalaforge, line 6: you should add your github handle here. that's your work 💪

tomalaforge (approved):

PR #998, opened 2024-06-12
Answer:53 signalStore approach

LMFinney:

## Warning:

### DIFF
--- a/apps/signal/53-big-signal-performance/src/app/user-form.component.ts
+++ b/apps/signal/53-big-signal-performance/src/app/user-form.component.ts
@@ -14,7 +14,7 @@ import { UserStore } from './user.service';
           class="rounded-md border border-gray-400"
           formControlName="name" />
       </div>
-      <div>
+      <div formGroupName="address">
         Address:
         <div>
           street:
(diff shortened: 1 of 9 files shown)


tomalaforge:
Nice, that's a nice use case for signal store if you are using signal store inside your application 👍

LMFinney:
> Nice, that's a nice use case for signal store if you are using signal store inside your application 👍

Thanks. I realized that the `withComputed` was unnecessary, so I just pushed a slightly simpler version.

apps/signal/53-big-signal-performance/src/app/user-form.component.ts
```diff
@@ -65,36 +65,26 @@ export class UserFormComponent {
...
-        ...u.address,
-        street: this.form.getRawValue().street,
-        zipCode: this.form.getRawValue().zipCode,
-        city: this.form.getRawValue().city,
-      },
-      note: this.form.getRawValue().note,
-      title: this.form.getRawValue().title,
-      salary: this.form.getRawValue().salary,
-    }));
+    this.userStore.update(this.form.getRawValue());
```
LMFinney, line 88: This gets a lot simpler when we use nested FormGroups, so that the shapes match.

PR #999, opened 2024-06-12
Answer:53 make everything primitive

LMFinney:
This is a non-signalStore implementation

## Warning:

### DIFF
--- a/apps/signal/53-big-signal-performance/src/app/user.service.ts
+++ b/apps/signal/53-big-signal-performance/src/app/user.service.ts
@@ -2,15 +2,11 @@ import { Injectable, signal } from '@angular/core';
 
 @Injectable({ providedIn: 'root' })
 export class UserStore {
-  user = signal({
-    name: 'Bob',
-    address: {
-      street: '',
-      zipCode: '',
-      city: '',
-    },
-    note: '',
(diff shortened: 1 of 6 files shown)


apps/signal/53-big-signal-performance/src/app/user.service.ts
```diff
@@ -2,15 +2,11 @@ import { Injectable, signal } from '@angular/core';
...
-    title: '',
-    salary: 0,
-  });
+  userName = signal('Bob');
+  userAddressStreet = signal('');
+  userAddressZipCode = signal('');
+  userAddressCity = signal('');
+  userNote = signal('');
+  userTitle = signal('');
+  userSalary = signal(0);
```
tomalaforge, line 11: This was the solution I was mostly looking for.
But for address, I will maybe not be so fine grained. most of the time, it's used together.

LMFinney: My initial implementation left address as an object, but then it was highlighted whenever anything else changed - Angular could tell that primitives hadn't changed, but it couldn't tell that a new address object was equal to the previous one.

Perhaps there's a more elegant solution, but this is what I went with.

PR #1001, opened 2024-06-12
Answer:53 Big Signal 

jdegand:
I thought about ways to mutate the object, but this has become significantly more difficult, now that `mutate` is gone.  I guess you could downgrade to Angular 16 and maybe that would work.

### DIFF
--- a/apps/signal/53-big-signal-performance/src/app/user.service.ts
+++ b/apps/signal/53-big-signal-performance/src/app/user.service.ts
@@ -1,4 +1,4 @@
-import { Injectable, signal } from '@angular/core';
+import { Injectable, computed, signal } from '@angular/core';
 
 @Injectable({ providedIn: 'root' })
 export class UserStore {
@@ -13,4 +13,12 @@ export class UserStore {
     title: '',
     salary: 0,
   });
+
+  readonly name = computed(() => this.user().name);
+  readonly note = computed(() => this.user().note);
+  readonly title = computed(() => this.user().title);
+  readonly salary = computed(() => this.user().salary);
+  readonly street = computed(() => this.user().address.street);
+  readonly zipCode = computed(() => this.user().address.zipCode);
+  readonly city = computed(() => this.user().address.city);
 }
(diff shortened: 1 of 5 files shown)


apps/signal/53-big-signal-performance/src/app/user.service.ts
```diff
@@ -13,4 +13,12 @@ export class UserStore {
...
     salary: 0,
   });
+
+  readonly name = computed(() => this.user().name);
+  readonly note = computed(() => this.user().note);
+  readonly title = computed(() => this.user().title);
+  readonly salary = computed(() => this.user().salary);
+  readonly street = computed(() => this.user().address.street);
+  readonly zipCode = computed(() => this.user().address.zipCode);
+  readonly city = computed(() => this.user().address.city);
```
tomalaforge, line 23: I haven't thought of this solution, but indeed that's a solution and I think it's how ngrx signal store is working under the hood 🔥

jdegand: My initial idea was to use `ngModel`, but that would have entailed a lot of code changes, so I figured there probably was a better way.

Signals are more commonly associated with template forms.  I don't know if you know about [ng-signal-forms](https://github.com/timdeschryver/ng-signal-forms) from Tim Deschryver, but it is worth a look for a possible future of signal forms.   This [video](https://www.youtube.com/watch?v=-yPnPZV3Bp8) makes a case against using signals with reactive forms.   

But this challenge is really about signals and objects, which led me to think about `mutate` and investigate ways to still be able to mutate signals.  This led me to discover [signalstory](https://github.com/zuriscript/signalstory), a new state management solution built around signals. It has mutate functionality. Also, `tan-stack` has mutations.    

One way to `mutate` and avoid change detection would be to use `return` inside `update`.  The object reference does not change and change detection will not run.   Another way to prevent change detection could be to use the fact that signals are not immutable.   You could do something similar to this: ` this.userStore.user().name = this.form.getRawValue().name` for all fields.  But these approaches are problematic when you need the other components to re-render.

tomalaforge: I know `ng-signal-form` but i have never used it. 
I'm waiting for the new signal form APIs of the angular framework that might come in the upcoming months (hopefully). 
But meanwhile, that's true that signals are not a good fit with reactive forms and better suited to template form.

PR #1009, opened 2024-06-18
Answer:54 pipe observable to signal

LMFinney:
Gives me memories of challenge 16 :)

## Warning:

### DIFF
--- a/apps/signal/54-pipe-observable-to-signal/src/app/currency.pipe.ts
+++ b/apps/signal/54-pipe-observable-to-signal/src/app/currency.pipe.ts
@@ -1,5 +1,4 @@
-import { inject, Pipe, PipeTransform } from '@angular/core';
-import { map, Observable } from 'rxjs';
+import { computed, inject, Pipe, PipeTransform, Signal } from '@angular/core';
 import { CurrencyService } from './currency.service';
 
 @Pipe({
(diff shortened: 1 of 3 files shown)


tomalaforge:
Nice work. Same solution as mine 🔥

LMFinney:
> Nice work. Same solution as mine 🔥

Thanks. My only concern with it is `allowSignalWrites`. I wish I knew a way to avoid it.

jdegand:
Haven't looked at or tested with your solution, but `allowedSignalWrites` can be avoided with `untracked`.

Check this [GitHub issue](https://github.com/angular/angular/issues/53986).

tomalaforge:
> Haven't looked at or tested with your solution, but `allowedSignalWrites` can be avoided with `untracked`.
> 
> Check this [GitHub issue](https://github.com/angular/angular/issues/53986).

Nice 🔥 I didn't know that
But it seems that `allowedSignalWrites` is going to be removed from the effect API

jdegand:
It seems like there will be multiple effects (the original effect and another for state changes or a root vs component effect distinction).   This might tie into an explicit tracking implemenation as well.

LMFinney:
> Haven't looked at or tested with your solution, but `allowedSignalWrites` can be avoided with `untracked`.
> 
> Check this [GitHub issue](https://github.com/angular/angular/issues/53986).

Thanks. I didn't know that, either. I've updated my PR to use that.

jdegand:
This [issue](https://github.com/angular/angular/issues/56407) also might be of interest.  

You can declare a signal inside a pipe then you don't have the `( signal | pipe )()` syntax in the template.  But it doesn't seem to apply in this challenge since the product is an input.

PR #1035, opened 2024-07-05
Answer:5

pietro2356:
I wanted to focus on implementing the interceptor for HTTP calls and the global error handler.

This is my first time using a global error handler and it was highly instructive to try to implement and use it.

Are there ways to better handle the direct conversion of observables to signals, or is it good as a solution?

As for the observables themselves, are there ways to clean them up once they are consumed and we get the results we wanted?

Thank you


ErickRodrCodes (approved):
Hello Pietro. This is a quite elegant solution! I was trying doing a simple one, but looking at your code, it is super clean and the separation of concerns is quite understandable. Great Job!

PR #1038, opened 2024-07-09
Answer:26

tsironis13:


tomalaforge:
nice job 👍 Do you use Nx at your work?

PR #1044, opened 2024-07-16
Answer:56. forms and signal

BillyOkoth:
feat: solution of reactive forms and signals
1. fetch the value from route params
2. convert the observable to signal
3. set the value to the form control


apps/signal/56-forms-and-signal/src/app/checkout.component.ts
```diff
@@ -15,6 +15,7 @@ import { products } from './products';
     <h2 class="mb-1 w-full bg-gray-400 p-2 text-white">Checkout</h2>
     <button
       routerLink="/order"
+      [queryParams]="{ quantity: quantity() }"
```
tomalaforge, line 18: why do you need this one?

apps/signal/56-forms-and-signal/src/app/order.component.ts
```diff
@@ -55,11 +57,18 @@ import { products } from './products';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export default class OrderComponent {
+  private route = inject(ActivatedRoute);
+
+  query$ = this.route.queryParams.pipe(map((q) => q['quantity'] as number));
+  productId = input('1');
```
tomalaforge, line 63: you can use input.required here,

apps/signal/56-forms-and-signal/src/app/order.component.ts
```diff
@@ -55,11 +57,18 @@ import { products } from './products';
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export default class OrderComponent {
+  private route = inject(ActivatedRoute);
+
+  query$ = this.route.queryParams.pipe(map((q) => q['quantity'] as number));
```
tomalaforge, line 62: you can use the snapshot version since you only want it when the component loads

package.json
```diff
@@ -34,6 +34,7 @@
     "@nx/angular": "19.1.0",
     "@swc/helpers": "0.5.3",
     "@tanstack/angular-query-experimental": "^5.12.1",
+    "node": "^18.18.2",
```
tomalaforge, line 37: why did you need to add node ?

apps/signal/56-forms-and-signal/src/app/checkout.component.ts
BillyOkoth: am passing the quantity data as part of the url in order to retrieve it on the order.component.ts  when we use the ActivatedRoute

package.json
BillyOkoth: I forgot to remove it since i had accidentally installed it

PR #1045, opened 2024-07-17
docs: update challenge count on main readme

jdegand:
I don't know if you'd want to change this to `50+` so you don't have to continually update this.


tomalaforge:
It should be done automatically with the nx generator. But there are some issues sometimes when two people create challenges at the same time.

tomalaforge (approved):

PR #1059, opened 2024-07-26
Answer:56

tsironis13:


apps/signal/56-forms-and-signal/src/app/order.component.ts
```diff
@@ -63,10 +64,24 @@ export default class OrderComponent {
...
+
+  constructor() {
+    effect(() => {
+      const q = this.quantity();
+
+      if (q) {
+        this.form.controls.quantity.patchValue(q);
+      }
+    });
+  }
```
tomalaforge, line 86: I did a similar solution using input signal. But I think I prefer the one of Billy https://github.com/tomalaforge/angular-challenges/pull/1044 using activatedRoute because the route is inialized before the component and you can initialize the form imediatly. Easier to understand and read.

PR #1080, opened 2024-08-08
Answer: 14

fpalmab:
Note that besides solving challenge 14: I added some extra configuration that Cypress was requesting to build the test.

- I stuck to using only rxjs for solving this challenge: I sent the information as a subject subscribed to the service instead of a value and added the async pipe on the end of the child component.


apps/rxjs/14-race-condition/src/app/app.component.ts
```diff
@@ -21,19 +23,22 @@ export class AppComponent implements OnInit {
   title = 'rxjs-race-condition';
   dialog = inject(MatDialog);
   topicService = inject(TopicService);
-  topics: TopicType[] = [];
+  topicSubject$ = new BehaviorSubject<TopicType[]>([]);
```
tomalaforge, line 26: This is one way of doing it, but you can do it in a declarative way. 

You can initiate topic directly inside your field initializer and avoid the ngOnInit hook

apps/rxjs/14-race-condition/src/app/app.component.ts
```diff
@@ -21,19 +23,22 @@ export class AppComponent implements OnInit {
...
+        this.topicSubject$.next(topics);
+        //this.dialog.closeAll();
+      });
   }
 
   openTopicModal() {
     this.dialog.open(TopicModalComponent, {
       data: {
-        topics: this.topics,
+        topics$: this.topicSubject$,
```
tomalaforge, line 41: I always prefer not to input observable, but this way it works

PR #1087, opened 2024-08-19
Answer: 2

fpalmab:
I'm not entirely sure I got the instructions correctly because I haven't used ngrx before.
My understanding of the assignment is that we should NOT be using selectors or effects if we are deriving the state from other selectors, so Status store shouldn't exist at all. Please correct me if that's not the objective. I was struggling with this one a lot.

-In my solution I moved the activity part of the component into its own.
-I removed the usage of the status selector and instead I'm filtering the activity using map and filter.


apps/ngrx/2-effect-vs-selector/src/app/store/user/user.service.ts
```diff
@@ -1,10 +1,16 @@
...
 
 @Injectable({
   providedIn: 'root',
 })
 export class UserService {
   fetchUser = () => timer(500).pipe(map(() => user));
+  updateUser = (user: User) =>
+    timer(1000).pipe(
+      map(() => {
+        return;
```
tomalaforge, line 13: ? 🤔

tomalaforge:
I must admin, I've very rusted on ngrx. I haven't used it recently and now i'm mostly using signals to save the state. So ngrx global store is not used very often. 

But the goal is to have only one source of trust which the store and you should derive the state from the store and not inside reducers.

PR #1093, opened 2024-08-26
Answer: 39

fpalmab:
Answering this challenge as a preamble to start solving challenge number 7

For this one, I created the InjectionToken with the identifier TIMER_VALUE and injected it into the Timer Component.

In order to avoid having to define my token value on each component that would want to implement it, I decided to set up a default provider value in my ApplicationConfig with a default value of 1000, so my video component remained loosely coupled to the timer component.

Afterwards, on the phone component I redefined the value of the same injector.

My only concern about using injectors over the Input directive is that it gives me the feel that it might be missed by people overlooking the code.


apps/angular/39-injection-token/src/app/app.config.ts
```diff
@@ -8,5 +9,6 @@ export const appConfig: ApplicationConfig = {
       { path: 'video', loadComponent: () => import('./video.component') },
       { path: 'phone', loadComponent: () => import('./phone.component') },
     ]),
+    { provide: TIMER_VALUE, useValue: DEFAULT_TIMER },
```
tomalaforge, line 12: you can do it even better and define your default value inside the cosntructor of your injectionToken

apps/angular/39-injection-token/src/app/timer.component.ts
```diff
@@ -11,5 +11,5 @@ import { DEFAULT_TIMER } from './data';
   `,
 })
 export class TimerComponent {
-  timer = toSignal(interval(DEFAULT_TIMER));
+  timer = toSignal(interval(inject(TIMER_VALUE)));
```
tomalaforge, line 14: 👍

apps/angular/39-injection-token/src/app/phone.component.ts
```diff
@@ -12,5 +13,6 @@ import { TimerContainerComponent } from './timer-container.component';
     </div>
     <timer-container />
   `,
+  providers: [{ provide: TIMER_VALUE, useValue: 2000 }],
```
tomalaforge, line 16: 👍

tomalaforge:
Very nice. you understand how DI works and that's a very good knowledge as an Angular developer. 

InjectionToken is very powerful, you will avoid prop drilling using input, the value will be store inside the DI and not as a component level. 

It might be missed but that's why the value is uppercase and developer should learn about it.

apps/angular/39-injection-token/src/app/app.config.ts
fpalmab: True. Thanks for the observation!

fpalmab: Wait, do you mean the Default Token value of the Default Provider Value? Because I didn't try just sending the provider without the useValue field because I thought it was required. 
Guess I need to re read how to set providers in Angular.

tomalaforge:
```ts
export const TIMER = new InjectionToken('call timer', {
  providedIn: 'root',
  factory: () => DEFAULT_TIMER,
});
```

PR #1107, opened 2024-09-03
Answer:8 Angular pure pipe

gitmbints:
Hello guys,

I hope you're all doing well. I’ve been working on the challenge n°8 (pure pipe), and I would greatly appreciate it if someone could take a moment to review my code.

Your feedback is invaluable, and I'm open to any suggestions or improvements.

Thank you in advance for your time and assistance!

Best regards,

Ambinintsoa

### DIFF
--- a/apps/angular/8-pure-pipe/src/app/heavy-computaion.pipe.ts
+++ b/apps/angular/8-pure-pipe/src/app/heavy-computaion.pipe.ts
@@ -1,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'heavyComputation',
+  standalone: true,
+})
+export class HeavyComputation implements PipeTransform {
+  transform(value: string, index: number): string {
+    return `${value} - ${index}`;
+  }
+}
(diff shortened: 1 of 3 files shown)


apps/angular/8-pure-pipe/src/app/heavy-computaion.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'heavyComputation',
+  standalone: true,
+})
+export class HeavyComputation implements PipeTransform {
+  transform(value: string, index: number): string {
+    return `${value} - ${index}`;
```
srikarphanikumar, line 9: Just to be on the safer side - can we add a null check here? Everything else is great!

Something like: 
````
transform(name: string | null | undefined, index: number | null | undefined): string {
    if (name == null || index == null) {
      return '';
    }
    
    return `${name} - ${index}`;
  }
  ````

gitmbints: Thanks for the review @srikarphanikumar.

PR #1112, opened 2024-09-06
docs: add Chinese translation

EnochGao:
add Chinese translation


tomalaforge:
oh waouhhhh. Thanks so much. I will not be able to review you 😅

tomalaforge (approved):

PR #1172, opened 2025-01-11
Answer:39 injection token

apalaio:


apps/angular/39-injection-token/src/app/data.ts
```diff
@@ -1 +1,8 @@
-export const DEFAULT_TIMER = 1000;
+import { InjectionToken, ValueProvider } from '@angular/core';
+
+export const TIMER = new InjectionToken<number>('default-timer');
```
tomalaforge, line 3: that's good, but you can be even better by using the factory function inside the `InjectionToken` class, so you can provide a default value and you only have to override it when needed.

PR #1173, opened 2025-01-14
Answer:16 master dependency injection

apalaio:


apps/angular/16-master-dependency-injection/src/app/app.component.ts
```diff
@@ -36,7 +59,7 @@ export class ProductDirective {
         </tr>
       </ng-template>
       <ng-template #body product let-product>
-        <tr>
+        <tr [currencyCode]="product.currencyCode">
```
tomalaforge, line 62: perfect

tomalaforge:
Be careful, you have two challenges inside this branch.

PR #1175, opened 2025-01-18
Answer:1

zealotrahl:


zealotrahl:
I can't get fully idea of ngTemplateOutlet, to me it looks too complicated; I believe this is straightforward and simple solution having the constraints mentioned;

tomalaforge:
Indeed your solution works, but if I make the exercise more complex like asking to add an icon for the city card only or a edit button or anything else, you will have a hard time and you will need to add a lot of condition inside your component.
Moreover you will need to edit your shared component which can be used by many teams and add a lot of friction. 
Your answer will work for this use case but it's not very scalable or maintainable in an application that will always require more features

apps/angular/1-projection/src/app/component/city-card/city-card.component.ts
```diff
@@ -1,12 +1,59 @@
...
+    this.http.fetchCities$.subscribe((e) => this.store.addAll(e));
+    this.store.cities$.subscribe((cities) => {
+      this.cityListItems = cities.map((city) => {
+        return {
+          name: city.name,
+          id: city.id,
+        };
+      });
+    });
+  }
```
tomalaforge, line 50: Don't use imperative coding, try to code in a declaration way, which means try to do the same but without the subscribe.

And use signals. 😉

apps/angular/1-projection/src/app/ui/card/card.component.ts
```diff
@@ -1,60 +1,39 @@
...
-      <button
-        class="rounded-sm border border-blue-500 bg-blue-300 p-2"
-        (click)="addNewItem()">
-        Add
-      </button>
-    </div>
+    <ng-content select="[card-img]"></ng-content>
+    <section>
+      <app-list-item
+        *ngFor="let item of list"
```
tomalaforge, line 12: no more ngfor, use the new control flow

tomalaforge:
Don't hesitate to ask question if something is unclear

PR #1177, opened 2025-01-20
Answer: 54 - Convert  pipe and observables to signals - 1

DzoeL123:
## Changes:

- feat: updated currency.service.ts methods and variables to use signals
- feat: updated transform methods to return price as string
- feat: removed unwanted use of async pipe in product-row.component.ts

## Warning:


tomalaforge:
Nice work, but it seems that version 19 is better with signal timing. 
Your solution will not work in Angular v18 😅, so that's a bit improvement on the Angular side.

To go a bit further, you could also refactor the `@Input`

PR #1225, opened 2025-02-11
Answer:40

Ian-Balijawa:
- Create web worker for prime number calculation
- Move heavy computation off main thread
- Add smooth progress updates via worker messages
- Improve UI feedback during calculation
- Add proper worker cleanup


tomalaforge:
be careful: you need to get the latest version
And you need to have only one challenge per PR.

For each challenge, start from the main branch, create a new branch and push that branch

PR #1263, opened 2025-03-27
Answer:58

tsironis13:
The solution is based on the fact that we cannot include 2 ng-content tags with the same selector. This is an Angular by design constraint. Ng-template tag can be used instead.


apps/angular/58-content-projection-condition/src/app/card.component.ts
```diff
@@ -1,25 +1,50 @@
...
+
+@Directive({
+  selector: '[appCardTitle]',
+})
+export class CardTitleDirective {}
+
+@Directive({
+  selector: '[appCardMessage]',
+})
+export class CardMessageDirective {}
```
tomalaforge, line 19: nice, using a directive to avoid magic string is the cherry on top. Nice solution

PR #1264, opened 2025-04-01
Answer: 59

StormLx:
Answer: 59


tomalaforge:
nice solution, I haven't thought of using the undefined status of httpResource. 

Thanks for the solution 🔥

PR #1267, opened 2025-04-02
Answer: 59

kevinagus:


apps/angular/59-content-projection-defer/src/app/page-2.ts
```diff
@@ -36,8 +42,22 @@ interface Post {
...
+    toObservable(this.triggerFetch)
+      .pipe(takeUntilDestroyed())
+      .subscribe((trigger) => {
+        if (trigger) {
+          this.postRessource = httpResource<Post[]>(
+            'https://jsonplaceholder.typicode.com/posts',
+            {
+              injector: this.#injector,
+            },
+          );
```
tomalaforge, line 58: very imperative, you could use a condition to trigger or not the request inside httpRessource.

apps/angular/59-content-projection-defer/src/app/expandable-card.ts
```diff
@@ -51,4 +57,10 @@ import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
 })
 export class ExpandableCard {
   public isExpanded = signal(false);
+  expanded = output<boolean>();
+  constructor() {
+    effect(() => {
+      this.expanded.emit(this.isExpanded());
```
tomalaforge, line 63: Using effect to trigger an output is often a sign of a bad design

kevin-agus: understood, I'll set a callback function for the button click instead

apps/angular/59-content-projection-defer/src/app/page-2.ts
kevin-agus: yep definitely I'll add a function to get a dynamic URL with a ternary condition

kevinagus: yep definitely I'll add a function to get a dynamic URL with a ternary condition

apps/angular/59-content-projection-defer/src/app/expandable-card.ts
kevinagus: understood, I'll set a callback function for the button click instead

PR #1280, opened 2025-04-04
Answer:59

tsironis13:
Second solution for challenge n59. 
The idea is to deligate the post http resource logic handling to dedicate directive.
The component calls the needed resource signal properties only when the card content is expanded.
By using such approach, using the conditional logic to trigger the http resource is not needed.

### DIFF
--- a/apps/angular/59-content-projection-defer/src/app/expandable-card.ts
+++ b/apps/angular/59-content-projection-defer/src/app/expandable-card.ts
@@ -1,4 +1,28 @@
-import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
+import { NgTemplateOutlet } from '@angular/common';
+import {
+  ChangeDetectionStrategy,
+  Component,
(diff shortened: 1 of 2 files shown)


apps/angular/59-content-projection-defer/src/app/expandable-card.ts
```diff
@@ -41,14 +66,29 @@ import { ChangeDetectionStrategy, Component, signal } from '@angular/core';
       class="overflow-hidden transition-[max-height] duration-500"
       [class.max-h-0]="!isExpanded()"
       [class.max-h-[1000px]]="isExpanded()">
-      <ng-content />
+      @defer (on interaction(expandContentButton)) {
```
tomalaforge, line 69: yes, yes and yes 
I was looking for this solution. There were a hint in the title. But I think I will create a second challenge for people to guess this solution. 

Great job. 🔥

PR #1305, opened 2025-04-17
Answer:50 - signal-bug-in-effect

HichamGuessab:
Answer:50

1. Need someone to know if I am well understanding and for deep dive question.
2. Any clue for the computed way ?

Thank you !


StormLx:
- I would use 3 separate effects too.
- I wouldn't use a ternary with undefined.
- The computed mustn't be used for side effects (such as showing an alert) but to computed a new value when a signal updates. So what I would do is a computed named "shouldDisplayAlert" that would calculate when we should show the alert, and use that computed in a single effect.

PR #1360, opened 2025-07-09
docs(async-redirect): fix author tag

jdegand:
The author links were missing in the README.  And I noticed the hyphen was missing so I think this is the reason why.


tomalaforge (approved):

tomalaforge:
thx

PR #1432, opened 2026-01-29
Answer: 1

Justin24506:
### Checklist:
- [x] refactor CardComponent and ListItemComponent 
- [x] keep `@for` in CardComponent
- [x] delete conditions in CardComponent
- [x] CSS: avoid using `::ng-deep`

### Description
Completed the first challenge using the Container/Presenter pattern and content projection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ package-lock.json
 npm-shrinkwrap.json
 
 __screenshots__/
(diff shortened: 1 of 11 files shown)


Justin24506:
I've just noticed that I've mistakenly mentioned Issue 1 in my commit. Sorry.

PR #1446, opened 2026-03-05
Answer: 60

tsironis13:
Not sure if `distinctUntilChanged()` is unnecessary inside the subscription.

### DIFF
--- a/apps/angular/60-async-redirect/src/app/routes.ts
+++ b/apps/angular/60-async-redirect/src/app/routes.ts
@@ -1,9 +1,27 @@
-import { Routes } from '@angular/router';
+import { inject } from '@angular/core';
+import { Router, Routes } from '@angular/router';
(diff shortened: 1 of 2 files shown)


apps/angular/60-async-redirect/src/app/routes.ts
```diff
@@ -1,9 +1,27 @@
...
 import { UserPage } from './user-page';
+import { UserProfileService } from './user-profile.service';
+
+const redirectTo = () => {
+  const router = inject(Router);
+
+  return inject(UserProfileService)
+    .getProfile()
+    .pipe(
+      distinctUntilChanged(),
```
tomalaforge: not sure that's useful, because the subscription is created on route navigation and removed after

tsironis13: thanks I will remove it

PR #1450, opened 2026-03-10
Answer: 62

tsironis13:


apps/forms/62-crossfield-validation-signal-form/src/app/app.component.ts
```diff
@@ -1,328 +1,104 @@
...
+        const confirmPassword = value();
+        const password = valueOf(schemaPath.password);
+        if (confirmPassword !== password) {
+          return {
+            kind: 'passwordMismatch',
+            message: 'Passwords do not match',
+          };
+        }
+        return null;
+      });
```
tomalaforge, line 64: 🔥

tomalaforge:
I can see that you don't like inline template 😂

PR #1451, opened 2026-03-12
Answer: 63

tsironis13:


apps/forms/63-child-forms/src/app/app.component.ts
```diff
@@ -1,288 +1,83 @@
...
-        validators: [Validators.required],
-      }),
-    });
-  }
+  protected checkoutForm = form(
+    this._checkoutModel,
+    (schemaPath) => {
+      buildAccountSection(schemaPath.account);
+      buildAddressSection(schemaPath.shipping);
+      buildAddressSection(schemaPath.billing);
```
tomalaforge, line 68: I didn't know you can apply the validation like this. I thought you had to call
```
apply(schema.shipping, buildAddressSection);
```

tomalaforge:
Thanks for your solution, I'm always learning new things. 🙏

PR #1454, opened 2026-03-14
Answer: 51

phhien203:
✅ Challenge Submission Checklist

Start your PR title with: Answer:${challenge_number}

⚠️ Important Notice

If you would like personal feedback or a detailed review, please support the project on GitHub:

- $5 — Review per PR
- $30 — Lifetime access to reviews and Q&A
- $100 — Lifetime access to reviews and Q&A + a 45-minute video call
  👉 https://github.com/sponsors/tomalaforge

You can also submit a PR without sponsorship to:

- Be listed among the answered challenges, or
- Receive a review from a community member. 🔥

### DIFF
--- a/apps/signal/51-function-call-effect/src/app/action.component.ts
+++ b/apps/signal/51-function-call-effect/src/app/action.component.ts
@@ -1,7 +1,6 @@
 import {
   ChangeDetectionStrategy,
   Component,
-  effect,
   inject,
   signal,
 } from '@angular/core';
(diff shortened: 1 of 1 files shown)


apps/signal/51-function-call-effect/src/app/action.component.ts
```diff
@@ -18,7 +17,8 @@ import { UserService } from './user.service';
       </label>
       <select
         name="actions"
-        [(ngModel)]="action"
+        [ngModel]="action"
+        (ngModelChange)="logMessage($event)"
```
tomalaforge, line 21: ahah, that's the better way but the goal was to use the untrack function. But that's a lot better. Need to find another use case 😅

phhien203: Yep. untrack is also nice. At least I learned how it is visualized in the Signal graph.

PR #1464, opened 2026-03-24
Answer:65

phhien203:
✅ Challenge Submission Checklist

Start your PR title with: Answer:${challenge_number}

⚠️ Important Notice

If you would like personal feedback or a detailed review, please support the project on GitHub:

- $5 — Review per PR
- $30 — Lifetime access to reviews and Q&A
- $100 — Lifetime access to reviews and Q&A + a 45-minute video call
  👉 https://github.com/sponsors/tomalaforge

You can also submit a PR without sponsorship to:

- Be listed among the answered challenges, or
- Receive a review from a community member. 🔥

### DIFF
--- a/apps/forms/65-signal-form-edition/src/app/user-form.component.ts
+++ b/apps/forms/65-signal-form-edition/src/app/user-form.component.ts
@@ -1,10 +1,9 @@
 import {
   ChangeDetectionStrategy,
   Component,
-  effect,
   inject,
   input,
-  signal,
+  linkedSignal,
 } from '@angular/core';
 import { rxResource } from '@angular/core/rxjs-interop';
 import { form, FormField, min, required, submit } from '@angular/forms/signals';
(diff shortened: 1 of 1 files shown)


apps/forms/65-signal-form-edition/src/app/user-form.component.ts
```diff
@@ -124,50 +134,51 @@ export class UserFormComponent {
...
 
   constructor() {
     effect(() => {
       const userValue = this.userResource.value();
       if (userValue) {
-        this.userForm.patchValue(userValue);
+        this.userModel.set(userValue);
       } else {
-        this.userForm.reset({ firstname: '', lastname: '', age: 0, grade: 0 });
+        this.userModel.set({ firstname: '', lastname: '', age: 0, grade: 0 });
```
tomalaforge: I created this challenge because you can avoid effect thanks to signal forms. 
You should look at linkedSignals. 

Everything should happen in the creation of userModel

phhien203: Thanks! I updated the code to use linkedSignal.

apps/forms/65-signal-form-edition/src/app/user-form.component.ts
```diff
@@ -124,50 +133,51 @@ export class UserFormComponent {
...
+      return rest;
+    } else {
+      return {
+        firstname: '',
+        lastname: '',
+        age: 0,
+        grade: 0,
+      };
+    }
+  });
```
tomalaforge, line 152: yes a lot nicer, isn't it

phhien203: Yep, that's right!

apps/forms/65-signal-form-edition/src/app/user-form.component.ts
```diff
@@ -124,50 +133,51 @@ export class UserFormComponent {
...
-    }),
-    grade: new FormControl(0, {
-      nonNullable: true,
-      validators: [Validators.required],
-    }),
+  userModel = linkedSignal<Omit<User, 'id'>>(() => {
+    const hasValue = this.userResource.hasValue();
+    const user = this.userResource.value();
+
+    if (hasValue && user) {
```
phhien203, line 140: Hi @tomalaforge , do you think it's ok to only check `if (user)` instead of `if (hasValue && user)`?

tomalaforge: I think, hasValue do !!user() so it's kind of the same

PR #1465, opened 2026-03-28
Answer:50

reed756:
I think the code originally doesn't work because the effect will run once when the first model changes and then will only track changes for that model that changed first, ignoring the others.

### DIFF
--- a/apps/signal/43-signal-input/src/app/user.component.ts
+++ b/apps/signal/43-signal-input/src/app/user.component.ts
@@ -2,8 +2,9 @@ import { TitleCasePipe } from '@angular/common';
 import {
   ChangeDetectionStrategy,
   Component,
-  Input,
-  OnChanges,
+  computed,
+  input,
+  numberAttribute,
 } from '@angular/core';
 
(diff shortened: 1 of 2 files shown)


tomalaforge:
Hello Ram, thanks for sponsoring the project.

have your read my latest article: https://medium.com/p/b0b69804496d
it's on this issue.

your fix might work but it's way to complicated and hard to understand. You can fix it in a very simpler way

reed756:
> Hello Ram, thanks for sponsoring the project.
> 
> have your read my latest article: https://medium.com/p/b0b69804496d it's on this issue.
> 
> your fix might work but it's way to complicated and hard to understand. You can fix it in a very simpler way

Hi @tomalaforge, thanks for the review. I've read the article and understand the fix. However, it will still trigger the alert every time you uncheck one of the boxes which doesn't match the acceptance criteria for the challenge. I couldn't think of another way to only trigger the alert when checking one of the boxes, other than having another signal to track the previous number of boxes that have been checked. Then you can compare the current number of boxes checked with the previous number to determine if you should show the alert or not.

PR #1473, opened 2026-04-12
Answer:53

lukasss88:


apps/signal/53-big-signal-performance/src/app/user-form.component.ts
```diff
@@ -64,36 +64,24 @@ export class UserFormComponent {
...
-    city: new FormControl(this.userStore.user().address.city, {
+    city: new FormControl(this.userStore.address.city(), {
       nonNullable: true,
     }),
-    note: new FormControl(this.userStore.user().note, { nonNullable: true }),
-    title: new FormControl(this.userStore.user().title, { nonNullable: true }),
-    salary: new FormControl(this.userStore.user().salary, {
+    note: new FormControl(this.userStore.note(), { nonNullable: true }),
+    title: new FormControl(this.userStore.title(), { nonNullable: true }),
+    salary: new FormControl(this.userStore.salary(), {
```
tomalaforge, line 79: nice, you could even transform this part and use the new signal form API 😊

